PR #241, opened 2015-02-25
default to in-memory option for SQLite in tests

dkj:


mgcam:
Tested npg_qc and SeqQC against this version. 
t/45-st-api-lims-traversal.t test fails - this is unrelated to the change, shoudl be fixed separately

PR #258, opened 2015-04-27
Optionally change group ownership of directories in analysis

jenniferliddle:


lib/Monitor/RunFolder/Staging.pm
```diff
@@ -217,6 +220,12 @@ sub move_to_analysis {
 
     move( $self->runfolder_path(), $destination );
 
+    my $group = get_config()->{group};
```
dkj: This needs a more descriptive config: use "analysis_group" within the already used "staging_areas", or something better....

lib/Monitor/RunFolder/Staging.pm
```diff
@@ -263,6 +272,23 @@ sub update_folder {
...
+    my $temp = $directory . '.original';
+    move($directory, $temp) || die "move error: $!";
+    mkdir $directory || die "mkdir error: $!";
+    for my $file (glob "$temp/*") {
+        my $dest = $directory . q{/} . basename($file);
+        move($file, $dest) || die "move($file, $dest)\nmove error: $!";
+    }
+    rmdir $temp || die "rmdir error: $!";
+
+    my $uid = getpwnam(getlogin());
```
dkj: How about -1 instead of working out the uid?

lib/Monitor/RunFolder/Staging.pm
```diff
@@ -263,6 +272,23 @@ sub update_folder {
     return;
 }
 
+sub _change_group {
+    my ($self, $group, $directory) = @_;
+    my $temp = $directory . '.original';
+    move($directory, $temp) || die "move error: $!";
```
dkj: "die" won't get past PerlCritic, nor will "$!"

PR #266, opened 2015-05-18
make sample name compalsory in samplesheet generation

mgcam:


mgcam:
At the moment this change is going to make matters worse for GCLP. Can be accepted when the db data are fixed.

PR #270, opened 2015-05-25
cope with variations in staging area prefixes

mgcam:


mgcam:
New location for status and staging area daemons is proposed. /nfs/sfXX/log directories should be created prior to code deployment.

PR #276, opened 2015-06-12
staging monitor extension

mgcam:


bin/staging_area_monitor
```diff
@@ -32,32 +39,49 @@ my @area_list = $monitor->validate_areas(@ARGV);
 
 croak 'No valid arguments' if scalar @area_list == 0;
 
-my $iteration = 0;
+# In production dev variable is not set
+my $dry_run = $ENV{'dev'};
```
dkj: Intentional loss of
 ```$ENV{dev} and ( $ENV{dev} ne 'live' )```
logic?

lib/Monitor/RunFolder/Staging.pm
```diff
@@ -275,8 +331,20 @@ sub _change_group {
 
     my $gid = getgrnam($group);
     chown -1, $gid, $directory;
+    # If needed, set group sticky bit
```
dkj: Terminology is non-obvious and I've got this wrong in the past. We're not using the "sticky" bit:
S_ISGID: take a new file's group from parent directory,
S_ISVTX: sticky bit (restricted deletion flag)

bin/staging_area_monitor
mgcam: > Intentional loss of
> |$ENV{dev} and ( $ENV{dev} ne 'live' )|
> logic?
>
Yes. This dry_run facility is, unfortunately, not clean, ie some tags 
are still set for the run. Therefore, if one wants to test the whole 
script, either more work has to be done or dev=dev still should be used. 
I was using it with dev=live and an early return so most of the code is 
not executed, but I still get true representation of the status of the 
file system in conjunction with the run status. Can put some extra time 
into setting tags conditionally, then we'll have a true dry_run mode




-- 

lib/Monitor/RunFolder/Staging.pm
mgcam: Added a better comment
> > +    # If needed, set group sticky bit



-- 

PR #283, opened 2015-06-30
Merge pull request #282 from wtsi-npg/devel

mgcam:
pull from wtsi devel to master to create release 84.6


mgcam:
wrong name/comment

PR #286, opened 2015-07-21
Dev no root 4 travis

jmtcsngr:


jmtcsngr:
The command "sudo apt-get install libgd2-xpm-dev" failed and exited with 1 during .

PR #297, opened 2015-09-03
Compositions

mgcam:


lib/npg_tracking/glossary/composition/factory.pm
```diff
@@ -0,0 +1,114 @@
...
+sub create_sequence_component {
+  my $self = shift;
+
+  my $class = 'npg_tracking::glossary::composition::component';
+  my $h = {};
+  for my $attr_obj ( $class->meta->get_all_attributes ) {
+    my $attr = $attr_obj->name;
+    $h->{$attr} = $self->can($attr) && defined $self->$attr ? $self->$attr : undef;
+  }
+  my $component;
```
dkj: Why are we avoiding the croak (should either position or id_run be missing) here?

dkj:
I'd like this to cope with non Illumina data as well - but rather have it merged in now than waiting on my nebulous specs.

PR #299, opened 2015-09-10
minor lims shim enhancement

dkj:
find driver_type from driver if provided

### DIFF
--- a/lib/st/api/lims.pm
+++ b/lib/st/api/lims.pm
@@ -17,7 +17,6 @@ with qw/  npg_tracking::glossary::run
        /;
 
 our $VERSION = '0';
-Readonly::Scalar our $LAST             => -1; #TODO - add -1 to critic's allowed values and purge codebase of constants equal to -1, or drop this todo
 
 =head1 NAME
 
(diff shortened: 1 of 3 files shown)


lib/st/api/lims.pm
```diff
@@ -120,6 +120,12 @@ has 'driver_type' => (
                      );
 sub _build_driver_type {
   my $self = shift;
+  if($self->has_driver){
```
mgcam: 1. Changes file update?
2. Driver attribute is of unknown type, it might be possible to set it to anything. Not clear why... Has to be fixed at some point.
Might be safer to have the following
if($self->has_driver && $self->driver) {
my $type = ref $self->driver;
if ($type) {
    my $prefix = __PACKAGE__ . q(::);
    $type =~ s/\A\Q$prefix\E//smx;
    return $type;
  }
}

dkj: 1. Had thought it was too minor for Changes - but will do. Makes for unambiguous procedure....
2. Code change is "safe" in so far as passing a none  "st::api::lims::" class object will report the whole class name, passing a HASH will report HASH, and passing a scalar or undef will report "". Restricting by type, or provided methods, the driver attribute could well be useful - desk neighbour was considering putting a DBIC object in there....

PR #304, opened 2015-10-19
st::api::lims new method any_sample_consent_withdrawn

keithj:
Create multi-attribute methods automatically.

    Added method st::api::lims::any_sample_consent_withdrawn.

    Removed tests of a private method.


mgcam:
New xml test data files to MANIFEST?

mgcam:
testes npg_ml_warehouse and data handling against this version

PR #310, opened 2015-11-26
factory to create composition from a list of rpt keys

mgcam:
Functionality to be used in library merge and later in a composite genotype check.


lib/npg_tracking/glossary/composition/component/illumina.pm
```diff
@@ -12,6 +12,16 @@ with qw( npg_tracking::glossary::composition::component
 
 our $VERSION = '0';
 
+sub filename {
+  my ($self, $ext) = @_;
+  $ext //= q[];
```
dkj: Do we need a minumum Perl version of 5.10 (Build.PL or code?) now we're using this operator? Not too fussed....

mgcam: > Do we need a minumum Perl version of 5.10 (Build.PL or code?) now 
> we're using this operator?
>
Requested at least perl 5.10 in Build.PL.




-- 

PR #319, opened 2016-02-01
Added 'purpose' to lims shim

jenniferliddle:


mgcam:
Having defaulted to 'qc' in the (old) wh driver and considering that the ml_warehouse driver might return undefined, it might be a good idea to create a wrapper for the 'purpose' call in st::api::lims to return 'standard' for a non-pooled lane or a plex whenever the driver returns undefined. Then the 'purpose' definition in st::api::lims::xml can be dropped.

PR #320, opened 2016-02-02
Enable UTF-8 support in SQLite test databases when it is enabled for MySQL

keithj:
Enable UTF-8 support in SQLite test databases when the appropriate attribute is set.

Check UTF-8 when reading/writing YAML.

Added an option to silence fixture loading chatter.


lib/npg_testing/db.pm
```diff
@@ -135,6 +183,11 @@ SQLite database. If the third argument is not given, a randomly
 named file in a temporary directory is used; this file will
 be cleaned up on exit.
 
+If UTF-8 is enabled for the MySQL database (using mysql_enable_utf8 =>
```
dkj: No longer correct?

PR #321, opened 2016-02-03
Enable UTF-8 support in SQLite test databases when it is enabled for MySQL.

keithj:
Check UTF-8 when reading/writing YAML.

Added an option to silence fixture loading chatter.


keithj:
This request is redundant now that the ClearPress workaround has been merged into devel.

PR #336, opened 2016-06-13
request mysql v5.7

mgcam:


mgcam:
5.7 is not available via Travis. More complex configuration is needed.

PR #338, opened 2016-07-12
immutable composition

mgcam:
…e (as much as

   it is practically possible in Perl):
     - functionality for an incremental assembly of a composition moved to
       a generic composition factory;
     - empty compositions are not allowed;
     - sort is performed by the constructor and removed from other functions;
     - integrity check when serializing composition;
 - Composition - find uses binary search.
 - Required version of List::MoreUtils increased to make binary search available.


mgcam:
dependency: https://github.com/wtsi-npg/npg_qc/pull/342

PR #339, opened 2016-07-13
lims code tidy-up

mgcam:
prior to moving to lims for compositions.

### DIFF
--- a/Build.PL
+++ b/Build.PL
@@ -211,6 +211,7 @@ my $builder = $class->new(
                 'MooseX::StrictConstructor'       => '0.08',
                 'namespace::autoclean'            => '0.09',
                 'Net::LDAP'                       => 0,
+                'open'                            => 0,
                 'Perl6::Slurp'                    => '0.03',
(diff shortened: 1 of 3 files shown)


Build.PL
```diff
@@ -211,6 +211,7 @@ my $builder = $class->new(
                 'MooseX::StrictConstructor'       => '0.08',
                 'namespace::autoclean'            => '0.09',
                 'Net::LDAP'                       => 0,
+                'open'                            => 0,
```
dkj: Is this required? i.e. is it not in core Perl?

mgcam: On 15/07/2016 09:41, dkj wrote:
> > +                'open'                            => 0,
>
> Is this required? i.e. is it not in core Perl?
>
It is core Perl. However, modules go in and out of core occasionally. 
Also, sometimes (not this case, I admit), one has to bump the version 
that is in core. And, I guess, it matters if one installs dependencies 
with an option to have latest even if a satisfactory version is already 
available. But ready to remove it if you think it would be better this way.


-- 

PR #349, opened 2016-09-01
Allow the samplesheet path to be determined by the environment

keithj:
These changes permit the samplesheet driver to find its samplesheet file path correctly from `NPG_CACHED_SAMPLESHEET_FILE` when necessary.

I moved the automatic setting of the path attribute from within the `st::api::lims::driver_type` builder  to its own builder method in `st::api::lims::samplesheet`.

I altered the tests to reflect that the `driver_type => 'samplesheet'` initarg alone may be used to set the driver type of `st::api::lims`, the samplesheet file `path` initarg no longer being required to pass to the driver's constructor.

### DIFF
--- a/lib/st/api/lims.pm
+++ b/lib/st/api/lims.pm
@@ -262,9 +262,8 @@ sub _build_driver_type {
     return $type;
   }
 
-  $self->_driver_arguments()->{'path'} ||= $ENV{$CACHED_SAMPLESHEET_FILE_VAR_NAME};
-  if ( $self->_driver_arguments()->{'path'} ) {
-    return $SAMPLESHEET_DRIVER_TYPE;
+  if ($ENV{$CACHED_SAMPLESHEET_FILE_VAR_NAME}) {
+    return 'samplesheet';
   }
 
   return $DEFAULT_DRIVER_TYPE;
--- a/lib/st/api/lims.pm
+++ b/lib/st/api/lims.pm
@@ -57,7 +57,7 @@ as this object's accessors. Example:
 
 =cut
 
-Readonly::Scalar my $CACHED_SAMPLESHEET_FILE_VAR_NAME => 'NPG_CACHED_SAMPLESHEET_FILE';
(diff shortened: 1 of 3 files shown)


lib/st/api/lims.pm
```diff
@@ -262,9 +262,8 @@ sub _build_driver_type {
     return $type;
   }
 
-  $self->_driver_arguments()->{'path'} ||= $ENV{$CACHED_SAMPLESHEET_FILE_VAR_NAME};
-  if ( $self->_driver_arguments()->{'path'} ) {
-    return $SAMPLESHEET_DRIVER_TYPE;
+  if ($ENV{$CACHED_SAMPLESHEET_FILE_VAR_NAME}) {
+    return 'samplesheet';
```
mgcam: return $SAMPLESHEET_DRIVER_TYPE; ?

lib/st/api/lims/samplesheet.pm
```diff
@@ -40,6 +40,8 @@ Readonly::Scalar  my  $NOT_INDEXED_FLAG             => q[NO_INDEX];
 Readonly::Scalar  my  $RECORD_SPLIT_LIMIT           => -1;
 Readonly::Scalar  my  $DATA_SECTION                 => q[Data];
 Readonly::Scalar  my  $HEADER_SECTION               => q[Header];
+Readonly::Scalar  my  $CACHED_SAMPLESHEET_FILE_VAR_NAME =>
+  q[NPG_CACHED_SAMPLESHEET_FILE];
```
mgcam: Similar definition in st::api::lims. Could we access that one?

keithj:
Returning `SAMPLESHEET_DRIVER_TYPE` is an improvement. Two definitions of `CACHED_SAMPLESHEET_FILE_VAR_NAME` is one definition too many. I feel that it should belong in `st::api::lims::samplesheet`. However, that's not possible because the package isn't always loaded when `st::api::lims` reads the variable, so I've made the change you suggest.

PR #358, opened 2016-10-19
Remove utilisation and XMP POST

mgcam:


mgcam:
Depends on https://github.com/wtsi-npg/npg_tracking/pull/357

PR #367, opened 2016-12-07
Add EBI Submissions link

dkj:


mgcam:
Tracking web server is not going to be updated in this deployment round. This will be tested withing a dev server for January release.

PR #371, opened 2017-01-27
remove unused module

mgcam:
This module is not used anywhere. The current notification system incorporated code fragments from this module. This code is easy to read, it gives better idea of what's going on than the currently used code.


mgcam:
All old code will be eventually removed

PR #372, opened 2017-01-27
text of notification e-mails updated and unified

mgcam:


mgcam:
@dkj, I suggest to remove reporting lane-level library name from all email notifications. For mlwh sourced LIMs data this name will be either undefined or, for non-pool lanes and lanes with one plex only, resolve to some id.

mgcam:
Not needed as such, superseeded by https://github.com/wtsi-npg/npg_tracking/pull/375

PR #373, opened 2017-02-01
test fixes to ensure a pass under Clearpress v. 475.1.20

mgcam:
Failure in http://cpansearch.perl.org/src/RPETTETT/ClearPress-v475.1.20/lib/ClearPress/view.pm,  line 294
  if(!$self->authorised) {
    #########
    # set http forbidden response code
    #
    $self->headers->header('Status', HTTP_FORBIDDEN);

- new in release 475.1.20

If the request processing goes through its normal lifecycle, the headers object is set by the time the render() method is called on the view object. In tests the view objects are created directly. This fix creates the view objects with the headers attribute set. The above code change only affects the tests that test authorisation failure.

### DIFF
--- a/t/20-view-administration.t
+++ b/t/20-view-administration.t
@@ -1,5 +1,6 @@
 use strict;
 use warnings;
+use HTTP::Headers;
 use English qw(-no_match_vars);
 use t::util;
 use npg::model::administration;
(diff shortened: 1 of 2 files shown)


mgcam:
Looked at the dev server with the latest Clearpress version. Fo some reason the value of one of HTTP headers is printed out in the top left corner of the screen... Enabling plan B - pinning Clearpress to the previous release, see https://github.com/wtsi-npg/npg_tracking/pulls/374. The tests will stay as they were so that their failure is a clear indication that the app is being deployed under the wrong version of Clearpress.

PR #380, opened 2017-03-06
Drop tables cost_code and cost_group from schema

jmtcsngr:


jmtcsngr:
Waiting for full set of changes to be ready

PR #390, opened 2017-04-11
tests fixes for gseq cluster

mgcam:
ensure /export -> /nfs


mgcam:
These changes fix problem on gseq - testes on gseq cluster against a new dated directory

PR #414, opened 2017-08-10
pick analysis using ref genome string

rube14:
Return aligner when parsing reference genome string
- This is a completely different approach from previous commit but the aim is the same: parse the _analysis_ name expected to be at the end of the reference genome string, between '[]', and even if there is no transcriptome version present (so not only for RNA-Seq)
- Change is compatible with current usage of the find() method in reference module
- Test module was changed to add new parsing scenarios
- Update the changelog accordingly

### DIFF
--- a/lib/npg_tracking/data/reference/find.pm
+++ b/lib/npg_tracking/data/reference/find.pm
@@ -366,30 +366,32 @@ sub lims2ref {
 
 Parses LIMs notation for reference genome, returns a list containing
 an organism, strain (genome version) and, optionally, a transcriptome
-version and/or an aligner; or an empty list.
+version and/or a word indicating the type of analysis to be run.
 
 =cut
 sub parse_reference_genome {
     my ($self, $reference_genome) = @_;
     $reference_genome ||= $self->reference_genome;
     if ($reference_genome) {
-        ##also allows for transcriptome version and aligner e.g. 'Homo_sapiens (1000Genomes_hs37d5 + ensembl_release_75) [star]'
-        ## no critic (Variables::ProhibitPunctuationVars RegularExpressions::ProhibitComplexRegexes)
-        my @array = $reference_genome  =~ qr{
-                                              (?<organism> \S+ ) \s+
-                                              [(] (?<strain> \S+ ) (?: \s+ \+ \s+ (?<tversion> \S+ ))? [)]
-                                              (?: \s+ [[] (?<aligner> \S+ ) []] )?
-                                          }smx;
-        if (scalar @array >= $THREE && $+{organism} && $+{strain}) {
-            if ($+{tversion} || $+{aligner}) {
-                ## assure list contents and order (aligner is 4th)
-                @array = ($+{organism}, $+{strain}, $+{tversion} // q[], $+{aligner} // q[]);
+        my ($organism, $strain, $tversion, $analysis, $parsed, @array);
+        ## allows for transcriptome version and analysis e.g. 'Homo_sapiens (1000Genomes_hs37d5 + ensembl_release_75) [star]'
+        $organism = '(?<organism>\S+)\s+';
+        $strain = '(?<strain>\S+)';
+        $tversion = '(?:\s+\+\s+(?<tversion>\S+))?';
+        $analysis = '(?:\s+[[](?<analysis>\S+)[]])?';
+        $parsed = $reference_genome  =~ qr{$organism [(] $strain $tversion [)] $analysis}smx;
+        $organism = $LAST_PAREN_MATCH{organism};
+        $strain = $LAST_PAREN_MATCH{strain};
+        $tversion = $LAST_PAREN_MATCH{tversion};
+        $analysis = $LAST_PAREN_MATCH{analysis};
+        if ($organism && $strain) {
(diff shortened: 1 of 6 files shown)


lib/npg_tracking/data/reference/find.pm
```diff
@@ -364,22 +365,33 @@ sub lims2ref {
...
-        if (! $a[2]) { $#a = 1; }
-        return @a;
-     }
-  }
-  return;
+    my ($self, $reference_genome) = @_;
+    $reference_genome ||= $self->reference_genome;
+    if ($reference_genome) {
+        ##also allows for transcriptome version and aligner e.g. 'Homo_sapiens (1000Genomes_hs37d5 + ensembl_release_75) [star]'
+        ## no critic (Variables::ProhibitPunctuationVars RegularExpressions::ProhibitComplexRegexes)
```
dkj: Can we rejig to avoid the no critic?

dkj: `$LAST_PAREN_MATCH{organism}` rather than `$+{organism}` will avoid triggering `Variables::ProhibitPunctuationVars` - though it is admitedly a little more verbose....

lib/npg_tracking/data/transcriptome/find.pm
```diff
@@ -62,21 +117,9 @@ has 'rnaseqc_gtf_file' => ( isa           => q{Maybe[Str]},
...
-  if (scalar @gtf_files == 0) {
-    if ($self->_organism_dir && -d $self->_organism_dir) {
-      $self->messages->push('Directory ' . $self->_organism_dir . ' exists, but GTF file not found');
-    }
-    return;
-  }
-  return $gtf_files[0];
+  $self->_set_subfolder('RNA-SeQC');
+  $self->_set_file_type('gtf');
+  return $self->_find_file;
```
mgcam: Reapeteadly resetting values for attributes _file_type and _subfolder might have unexpected side effects. It would be cleaner for the '_find_file' method to accept file type and subfolder as input.

lib/npg_tracking/data/transcriptome/find.pm
```diff
@@ -2,13 +2,43 @@ package npg_tracking::data::transcriptome::find;
...
+has '_index_name_ext' => (isa      => q{Str},
+                          is       => q{ro},
+                          required => 0,
+                          lazy     => 1,
+                          builder  => q{_build_index_name_ext},);
+
+sub _build_index_name_ext {
+    my $self = shift;
+    my $ext = $self->analysis eq $DEFAULT_ALIGNER ? q[bt2] :
+              $self->analysis eq $DEFAULT_QUANTIFIER ? q[json] : return;
```
mgcam: Would be cleaner as 
my $ext = $self->analysis eq $DEFAULT_ALIGNER ? q[bt2] :
      ($self->analysis eq $DEFAULT_QUANTIFIER ? q[json] : undef);

lib/npg_tracking/data/reference/find.pm
```diff
@@ -364,22 +365,35 @@ sub lims2ref {
...
+        ## allows for transcriptome version and analysis e.g. 'Homo_sapiens (1000Genomes_hs37d5 + ensembl_release_75) [star]'
+        $organism = '(?<organism>\S+)\s+';
+        $strain = '(?<strain>\S+)';
+        $tversion = '(?:\s+\+\s+(?<tversion>\S+))?';
+        $analysis = '(?:\s+[[](?<analysis>\S+)[]])?';
+        $parsed = $reference_genome  =~ qr{$organism [(] $strain $tversion [)] $analysis}smx;
+        $organism = $LAST_PAREN_MATCH{organism};
+        $strain = $LAST_PAREN_MATCH{strain};
+        $tversion = $LAST_PAREN_MATCH{tversion};
+        $analysis = $LAST_PAREN_MATCH{analysis};
```
mgcam: Quoted around hash keys will emphasise that these are strings rather than function names, the code will run faster.

lib/npg_tracking/data/transcriptome.pm
```diff
@@ -1,13 +1,15 @@
 package npg_tracking::data::transcriptome;
 
 use Moose;
+use Readonly;
```
mgcam: This package is not used

lib/npg_tracking/data/transcriptome/find.pm
```diff
@@ -2,13 +2,34 @@ package npg_tracking::data::transcriptome::find;
...
+has '_index_name_ext' => (isa      => q{Maybe[Str]},
+                          is       => q{ro},
+                          required => 0,
+                          lazy     => 1,
+                          builder  => q{_build_index_name_ext},);
+
+sub _build_index_name_ext {
+    my $self = shift;
+    ## which type of file to look for in the index folder based on the analysis
+    my $ext = $self->analysis eq $DEFAULT_ALIGNER ? q[bt2] :
```
dkj: There maybe an opportunity to improve the semantics here and also w.r.t. using `DEFAULT_`

dkj:
I think the analysis attribute may need to move to a different, more common, module in the future.

PR #425, opened 2017-10-05
Typed rptkey inflation

mgcam:
Tested npg_qc, npg_seq_melt, npg_seq_pipeline, npg_ml_warehouse against this change. Two tests in npg_qc will have to be adjusted (error message for an invalid rtp key has changed).


jmtcsngr:
Testing with regex:

https://regex101.com/r/5tN2bj/2

PR #426, opened 2017-10-10
cope with local soft links of files e.g. in transcriptome directories

dkj:
Don't bail out on the existence of multiple files if they are in fact the same (via symlinks).


dkj:
Alternatively, fix our data so this isn't required....

PR #430, opened 2017-10-27
remove uptime info from instrument page

dkj:
function no longer available


mgcam:
How did I miss this?.. Thank you, @dkj

PR #457, opened 2018-05-30
Use CORS support in xml feeds

jmtcsngr:


mgcam:
We can get rid of urls.tt2. The only line from this template can be moved to actions.tt2

PR #461, opened 2018-06-04
Added script to call Monitor::RunFolder::Staging->move_to_analysis()

jenniferliddle:


bin/npg_move_runfolder
```diff
@@ -0,0 +1,75 @@
...
+
+=head1 SYNOPSIS
+
+    C<npg_move_runfolder --runfolder_path path>
+
+=head1 DESCRIPTION
+
+This script moves 'incoming' to 'analysis' for a given runfolder,
+and changes the group name to that found in the staging configuration.
+
```
mgcam: The description seems strange. Should read "This script moves a given run folder from 'incoming' directory to 'analysis' directory. "group name to that found in the staging configuration" is rather criptic.

bin/npg_move_runfolder
```diff
@@ -0,0 +1,75 @@
...
+use Monitor::Staging;
+use Monitor::RunFolder::Staging;
+
+our $VERSION = '0';
+
+use Readonly;
+
+my $mon = Monitor::RunFolder::Staging->new_with_options(status_update => 0);
+my @ms = $mon->move_to_analysis();
```
mgcam: It is still possible to overwrite status_update from the command line. Do we want this?

mgcam:
The new script should be added to MANIFEST. The change should be added to Changes file. There should be a test for no status update

bin/npg_move_runfolder
dkj: I'm happy to keep that "free" functionality.

PR #462, opened 2018-06-05
change aligner attr to lazy_build

mgcam:
so that child classes could set their own defaults


mgcam:
Not essential, closing for now

PR #482, opened 2018-10-08
extension of the serialization framework for composition/component

mgcam:


keithj:
There's a Critic fail which is just trailing whitespace in `npg_tracking::glossary::composition::serializable`

PR #485, opened 2018-10-11
Tileviz link

mgcam:


mgcam:
See tileviz index creation in https://github.com/wtsi-npg/npg_seq_pipeline/pull/377

PR #489, opened 2018-11-13
use file system type to switch inotify option on/off

mgcam:


mgcam:
Tested both on seq farm and OpenStack with lustre fs mounted

mgcam:
It's important to check that the argument path exists since '/' mount point is returned for non-existing paths

PR #491, opened 2018-11-23
Add support for (cD|R)NA library types in npg_tracking::data::snv::find

rube14:
- Add tests and some auxiliary files for LIMS data


lib/npg_tracking/data/bait/find.pm
```diff
@@ -16,12 +16,19 @@ with qw/ npg_tracking::data::reference::find /;
...
+Readonly::Scalar my $DEFAULT_BAIT_NAME => q{Standard};
+Readonly::Scalar my $RNA_LIBRARY_BAIT_NAME => q{Exome};
+
 
 has 'bait_name'     => ( isa => q{Maybe[Str]}, is => q{ro}, lazy_build => 1,
                         documentation => 'Bate name',);
 sub _build_bait_name {
   my $self = shift;
-  return $self->lims->bait_name;
+  my $bait_name = $self->lims->bait_name // $DEFAULT_BAIT_NAME;
```
mgcam: Using default on line 27 changes the previous behaviour of the code and can potentially break lots of code in the pipeline. Why is it needed?

rube14: A better approach has been considered. Thanks for the input.

rube14:
PR depended on by https://github.com/wtsi-npg/npg_qc/pull/582

PR #493, opened 2018-12-18
drop fancy alien to DBIx status-related code

mgcam:


lib/npg_tracking/Schema/Result/Instrument.pm
```diff
@@ -410,12 +395,11 @@ Change the instrument_status for an instrument. This means adding a new row to
...
-description or username fields respectively.
+Two arguments are required, an instrument status description and a user id or
+name.
 
-    $instrument_result_object->update_instrument_status( 4, 'jo3' );
-    $instrument_result_object->update_instrument_status( 'Down', 5 );
+    $instrument_result_object->update_instrument_status( 'Down', 'jo3');
+    $instrument_result_object->update_instrument_status( 'Down', '5);
 
 =cut
```
jmtcsngr: I think status_description (instead of status_identifier) is better name for the parameter for this function.

lib/npg_tracking/Schema/Result/RunStatusDict.pm
```diff
@@ -122,71 +136,42 @@ our $VERSION = '0';
...
+sub compare_to_status_description {
+  my ($self, $status_desc) = @_;
+
+  $status_desc or croak 'Non-empty status description string required';
+  my $other_status_row = $self->result_source()->resultset()
+                         ->find({description => $status_desc});
+  if (!$other_status_row) {
+    croak "Run status description '$status_desc' does not exist";
+  }
+  my $this_ti = $self->temporal_index;
```
jmtcsngr: This can be moved earlier so we only try to find the other status row when we have checked this row has temporal index.

PR #513, opened 2019-04-29
Delete runs from DR servers

jenniferliddle:


bin/npg_delete_dr_runs
```diff
@@ -14,51 +14,98 @@ use npg_tracking::Schema;
...
 while (my $run=$rs->next()) {
     my $status = $run->current_run_status_description || "";
     my $tag = $run->is_tag_set("staging");
     my $run_folder = $run->folder_name || "";
     if ($run_folder and !$tag ) {
-        print $run_folder, "\t", $run->id_run, "\t", $run->folder_path_glob, "\t", $status, "\n";
+        my $prefix = substr($run_folder,0,4);
+        my $dname = "/exports/$prefix/$run_folder";
+        if ($files and !$folders) {
+            $cmd .= "if [ -d $dname ]\\\; then rm $dname/Data/Intensities/BaseCalls/L\*/C\*/\*.cbcl\\\; fi\\\;";
```
dkj: I think we should be able to drop the second level of escaping

bin/npg_delete_dr_runs
```diff
@@ -14,51 +14,98 @@ use npg_tracking::Schema;
...
+            $cmd .= "if [ -d $dname ]\\\; then rm -r $dname\\\; fi\\\;";
+        }
+        if (!$files and !$folders) {
+            $cmd .= "if [ -d $dname ]\\\; then ls -ld $dname\\\; fi\\\;";
+        }
     }
 }
+chop $cmd;
+if ($no_dry_run) {
+    system("echo $cmd | ssh ukbb\@seq-stor-01 bash");
```
dkj: We could open filehandles for each `ssh` process forked and write commands run folder by runfolder to each filehandle. Printing to those avoids the `echo` and extra level of escaping. 
I fear `echo $cmd` might grow too large for the shell.
What about running `bash -e` as we're not expecting failures given the testing for directories.

mgcam: There is more to closing filehandles like this than is currently in the code, see https://perldoc.perl.org/functions/close.html

bin/npg_delete_dr_runs
```diff
@@ -0,0 +1,170 @@
...
+our $VERSION = '0';
+
+my $help;
+my $files;
+my $folders;
+my $no_dry_run;
+
+sub send_commands {
+    my ($dr_server, $cmd) = @_;
+    open(BASH, "| ssh $dr_server bash -e") or die("Can't run ssh: $!");
```
dkj: Would prefer the `open FILEHANDLE,MODE,EXPR,LIST` form
We use `croak` elsewhere in this script.
```suggestion
    open(BASH, '|-', 'ssh', $dr_server, qw(bash -e)) or croak("Can't run ssh: $!");
```
(untested)

bin/npg_delete_dr_runs
```diff
@@ -0,0 +1,170 @@
...
+pod2usage(0) if $help;
+
+my $rs=npg_tracking::Schema->connect()->resultset(q(RunStatusDict))
+    ->search({description=>[q[run cancelled],q[data discarded],q[qc complete]]})
+    ->related_resultset(q(run_statuses))->search({q(run_statuses.iscurrent)=>1})
+    ->related_resultset("run")->search({q(run.folder_path_glob)=>{LIKE=>q(%/esa-%)}});
+
+my $cmd = '';
+while (my $run=$rs->next()) {
+    my $status = $run->current_run_status_description || "";
```
dkj: unused
```suggestion
```

bin/npg_delete_dr_runs
```diff
@@ -0,0 +1,170 @@
...
+            $cmd .= "if [ -d $dname ]; then rm -r $dname; fi\n";
+        }
+        if (!$files and !$folders) {
+            $cmd .= "if [ -d $dname ]; then ls -ld $dname; fi\n";
+        }
+    }
+}
+chop $cmd;
+if ($no_dry_run) {
+    send_commands('ukbb\@seq-stor-01', $cmd);
```
dkj: Uncomfortable with hostnames hardcoded (in this repo) - if staying in `npg_tracking` these should be able to be set from command line options (but they could be defaults - but include as fqdn if so).

dkj:
We could probably drop the cbcl only option now.

dkj:
We should reconsider integrating the hard linking for data products and cached top-up candidates.

bin/npg_delete_dr_runs
jenniferliddle: Simplifying things is good. I like simple.

jenniferliddle:
> We could probably drop the cbcl only option now.

We _could_,  but I'd be inclined to keep the option in for now at least. There is nothing to say we have to use it.

jenniferliddle:
> We should reconsider integrating the hard linking for data products and cached top-up candidates.

Sounds good. Do we have code for hard linking for data products and cached top-up candidates to integrate? Where is it?

jenniferliddle:
Being re-written and moving to another repository.

PR #517, opened 2019-05-15
samplesheet for multiple runs

mgcam:


jidur:
This branch of the code worked fine using a multi-run sample sheet (sample sheets from 2 runs combined with id_run added as an additional field at the end and the header from the 2nd run removed) for a verify bam id and bcfstats job which had previously been unable to run on the ESA.

PR #519, opened 2019-05-16
Samplesheet generation with data from multiple runs

mgcam:
Based on #517

Example of generating a samplesheet:

perl -le 'use npg::samplesheet; use st::api::lims; my $l=st::api::lims->new(rpt_list=>"26480:1:9;26480:2:9;26480:3:9;26480:4:9;28780:2:4",driver_type=>"ml_warehouse_auto");npg::samplesheet->new(extend => 1, lims =>[$l->children], output =>"my_samplesheet.csv")->process'


lib/st/api/lims/samplesheet.pm
```diff
@@ -269,11 +232,120 @@ sub _build_spiked_phix_tag_index {
...
+  }
+  if (defined $self->tag_index) {
+    $s .= q[ tag_index ] . $self->tag_index;
+  }
+  return $s;
+}
+
+#####
+#
+# Dinamically create standard driver methods as defined in st::api::lims
```
jidur: typo : dynamically

PR #546, opened 2019-12-13
remove redundant script

mgcam:
... since a move to analysis is done by staging daemon


mgcam:
Functionality is still potentially in use

PR #547, opened 2019-12-13
removed ability to inspect run folders via ftp to instrument's computer

mgcam:
See Changes file for detailed description.


dkj:
Will the daemon stay up if it hits a problematic runfolder with: 739b52335eecd07df63fdfdcdc812eb8e058547c ?

mgcam:
> Will the daemon stay up if it hits a problematic runfolder with: [739b523](https://github.com/wtsi-npg/npg_tracking/commit/739b52335eecd07df63fdfdcdc812eb8e058547c) ?
Yes, it will because the code for considering a single run folder is inside the try-catch. The top-level try-catch was one of later additions on request by @srl147and no proper review was done at the time.

PR #579, opened 2020-06-18
Support additional revision field in primer panel name

ces:


Changes
```diff
@@ -1,5 +1,7 @@
 LIST OF CHANGES
 
+ - Optional revision filed added to primer panel name in LIMs.
```
mgcam, line 3: filed -> field?

ces: Yes :)

dkj:
Thoughts on trimming off the revision when looking up in repository?
(I'd thought to alter what's in the repository to avoid code change, but if we're forced to make a code change perhaps we should just look up on the first 2 segments and ignore the 3rd)

ces:
> Thoughts on trimming off the revision when looking up in repository?
> (I'd thought to alter what's in the repository to avoid code change, but if we're forced to make a code change perhaps we should just look up on the first 2 segments and ignore the 3rd)

I did ask about this.. but yes sure happy to ignore third if the convention will be revisions will always have same coordinate. (Yes obviously it would have been easiest if they called it V3_B rather than V3/B as that would have just been symlink...).

dkj:
@ces apologies: indecisiveness on my part. The extra code required, which you've been able to purge, knocked me off the fence!

PR #587, opened 2020-11-25
Novaseq v1.5

srl147:


MANIFEST
```diff
@@ -969,9 +969,12 @@ t/data/run_info/runInfo.miseq.xml
 t/data/run_info/runInfo.nextseq.xml
 t/data/run_info/runInfo.novaseq.xml
 t/data/run_info/runInfo.novaseq.xp.xml
+t/data/run_info/runInfo.novaseq.v1.5.xp.xml
```
dkj, line 972:
```suggestion
t/data/run_info/runInfo.novaseq.xp.v1.5.xml
```

dkj: (My Travis build hung because of this I think...)
Might be worth a rebase (fixing conflict) on top of current devel HEAD.

PR #589, opened 2020-11-26
obscure, not used and/or broken functionality removed

mgcam:


mgcam:
took away code for instruments' modifications, which is needed

PR #602, opened 2021-02-25
speed-up (de)serialization for compositions

mgcam:


nerdstrike (changes requested):
Just a few typos to tidy up

nerdstrike (approved):

PR #609, opened 2021-03-16
remove unused views for an instrument manufacturer

mgcam:


mksanger (commented):
Do we need a separate `manufacturer` class, could it be removed in favour of just having a `manufacturer_name` method in `instrument_format`?  That appears to be the only use of that class after these changes.

mksanger (approved):
Tested with an openstack machine.

PR #615, opened 2021-06-10
Optionally save run and lane statuses to the database.

mgcam:
The status monitor is going to be decomissioned in favour of saving
statuses to the database by pipeline jobs. We'd like to maintain an
ability to cteate JSON files with statuses, so the npg_status2file script
is extended to optionally save the status to a database.

When the pipeline is changed to use the new `--db_save` option, we might need to temporarily comment out JSON file creation so that the status daemon, which will have to run for some time after the change is introduced is not creating  superfluous database entries.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,8 @@
 LIST OF CHANGES
 
+ - Extend npg_status2file script to optionally save statuses to the tracking
+   database
+ 
 release 91.12.0
  - Change instrument and instrument format pages to only show current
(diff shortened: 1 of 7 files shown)


nerdstrike (changes requested):
I don't see any problems here. Wouldn't hurt to put the script rename into the changelog.

nerdstrike (approved):

PR #617, opened 2021-07-07
Remove live tests for st::api::lims XML driver.

mgcam:


nerdstrike (approved):
No objections from me to making this more simple

PR #618, opened 2021-07-09
Custom location of the Clearpress configuration file for unit tests.

mgcam:
NPG_TEST_TRACKING_CONFIGPATH environment variable can be used to set
a custom location of the Clearpress configuration file for unit tests.
If this variable is unset, a default location of the configuration
is used.

This functionality will be used when running a full stack tests when we do not have a direct access to tracking conf file for tests. Password-less MySQL access for a root is not encouraged these days.


nerdstrike (approved):
Confirmed to work

dkj (approved):
(code read only)

PR #624, opened 2021-08-10
Add no_auto, no_auto_analysis, no_auto_archive tags to tag table

jenniferliddle:
Required by recent pipeline changes https://github.com/wtsi-npg/npg_seq_pipeline/pull/677/


mgcam:
As far as I know, pre-setting tags is not required.

mgcam:
Adding a tag to the table does not result in this tag appearing in the list of available tags when setting tags. Just from playing with the web page, the tag should be used an least once to make it appear in the list. Therefore, their is no advantage in pre-setting these tags.

PR #625, opened 2021-08-11
Remove the API wrapping around tracking XML API.

mgcam:
This API has not been used in production for a long time. The part of
it that links run id and batch id (npg::api::run) was used in tests.
Tests are amended either to use samplesheets is a source of LIMS data
or, where tests are still using XML LIMS driver for st::api::lims,
batch id is used explicitly.

st::api::lims::xml previously had dependency on npg::api::run to allow for retrieval of batch_id
from run XML API in cases where it was not supplied explicitly. This dependency is removed without
changing the public interface of this class. 

Before this pr is merged, npg_qc and npg_seq_warehouse tests (if they still exist by that time)
should be run against this branch to see if any tests fail. If failed tests are found, they have to be updated
in the same way the tests in this package were updated.  All other tests should be run as well - just in case.

https://github.com/wtsi-npg/npg_qc/pull/781 should be merged first


nerdstrike (approved):
Hooray for deleting code!

PR #642, opened 2021-11-02
Create a script to make it easier to add groups for users

nerdstrike:
Is this what you had in mind @mgcam ? Or do you need even more admin capabilities?

### DIFF
--- a/scripts/change_user_roles.pl
+++ b/scripts/change_user_roles.pl
@@ -46,7 +46,6 @@ $schema->txn_do(
                 print "Assigned $new_role to user\n";
             }
         }
-        die "You shall not pass!";
     }
 );
 
--- a/scripts/change_user_roles.pl
+++ b/scripts/change_user_roles.pl
@@ -6,48 +6,76 @@ use Pod::Usage;
 
 use npg_tracking::Schema;
 
-my $username = 'kt19';
+my $username;
+my $help;
+my $list;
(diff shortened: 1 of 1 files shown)


jenniferliddle:
At a minimum I would suggest adding -h, -? and --help options together with a 'Usage' display.

'Nice to have' would be options to display a list of valid roles, and existing roles for a given user.

nerdstrike:
To be fair, it probably shouldn't set my role by default either... Tweaks will be forthcoming

nerdstrike:
@jenniferliddle - is that more to your liking?

jenniferliddle:
> @jenniferliddle - is that more to your liking?

Yay! That looks good :-)

The only problem I have is that when I try to run it I get an error:

> DBIx::Class::Storage::DBI::_dbh_execute(): DBI Exception: DBD::mysql::st execute failed: INSERT command denied to user 'npgro'@'seqfarm3-head2.internal.sanger.ac.uk'

Is that just because I have an odd .npg/npg_tracking-Schema file?

nerdstrike:
I think that's a fairly straightforward issue. You do what you've got to do to get `npgrw` as your database user. The --list options should still work for `npgro` but... that's the least exciting bit. The script is leaning on the "npg" way of authenticating, so it can't do anything about picking your tracking DB user.

PR #650, opened 2021-12-06
Add folder name to run listing pages.

absanger:
Resolves: #647


mgcam (changes requested):
Folder name might be undefined, please use non-breaking space in this case

PR #652, opened 2021-12-15
Remove old team names from 'add run' dropdown

mksanger:
Remove options 'B', 'C' and 'joint'.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,8 +1,9 @@
 LIST OF CHANGES
 
+ - Removed old team names from run add template
  - Made NovaSeq instruments skip wash statuses after completing a run, this
    does not affect statuses set manually through the UI.
- 
+
 release 92.2.0
  - Added singularity definition for the tracking web server.
  - Tracking web pages:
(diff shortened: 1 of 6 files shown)


mksanger:
Resolves [RT736224](https://rt.sanger.ac.uk/Ticket/Display.html?id=736224)

mgcam (changes requested):
Incorrect team validation against the redundant list of teams.
The old list of teams is no longer needed in the run model.

mksanger:
Don't we need to be able to validate old team names so that we can display old runs?

mgcam:
On 06/01/2022 10:39, mksanger wrote:
>
> Don't we need to be able to validate old team names so that we can 
> display old runs?
>
As far as I can see, the validation is only performed when a new run is 
created.
>
> Message ID: ***@***.***>
>


-- 
 The Wellcome Sanger Institute is operated by Genome Research 

mgcam (approved):

PR #654, opened 2021-12-17
Make NovaSeq instruments skip `wash required` status

mksanger:
As requested in [RT737813](https://rt.sanger.ac.uk/Ticket/Display.html?id=737813).

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,7 @@
 LIST OF CHANGES
 
+ - Made NovaSeq instruments skip 'wash required' status.
+ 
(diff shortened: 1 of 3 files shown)


mgcam (changes requested):
The user has clarified his request. It is now clear that the auto-change should be straight to the 'up' instrument status. Therefore, the code is correct.
The entry in the Changes file should make clear that this change does not affect instrument statuses assigned via the UI.

The test is set incorrectly.

mgcam (approved):

PR #659, opened 2022-01-10
Change default lims driver to samplesheet

mgcam:


mgcam:
Marked as work in progress so that it is not merged too early. We are going to have a release of web-server changes first.

The change in this pr requires a full stack test.

mgcam:
Tested within a full tack deployment

PR #660, opened 2022-01-13
remove collocation checks for staging area links

jmtcsngr:
After moving npg_tracking to containers in head nodes and once staging areas
are not local to head nodes, the check is not valid anymore. We expect staging
areas will be independent of farm head nodes (different filesystem) and we are
not allowing the container to access local filesystems where staging areas
would be.

### DIFF
--- a/Build.PL
+++ b/Build.PL
@@ -210,7 +210,6 @@ my $builder = $class->new(
                 'Template'                        => '2.19',
                 'Try::Tiny'                       => 0,
                 'URI::Escape'                     => 0,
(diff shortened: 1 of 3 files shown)


mgcam:
The comment to now deleted code states "It is important that the development servers serve all urls themselves and do not delegate some urls to production servers." The redirection from the dev servers will not go to the production servers. This should not block this pr. In future the code might be refactored to ensure that the port of the current URL is propagated to the port of the URL used for redirection.

mgcam (approved):
This change has been tested on live server.

PR #669, opened 2022-02-10
Validate id run

srl147:


mgcam (changes requested):
We should be setting run folder name for run sin as many statuses as we can up to the point when run reaches 'analysis pending', otherwise we will will be unable to automatically delete cancelled and similar runs. Temporal indexes and run state comparison should be used for this purpose. We should not be updating very old runs, ie whenthe current status of the run is well in the past, say, over a week old.

lib/Monitor/Staging.pm
```diff
@@ -91,9 +91,14 @@ sub find_live {
             } else {
                 if (! $run_row->folder_name ) {
                   warn qq[Folder name in db not available, will try to update using '$run_folder'.];
+                  # check the id_run corresponds to a run with a status of run pending
```
dkj, line 94:
```suggestion
                  # check the id_run corresponds to a run with a status of "run pending" to reduce the
                  # chance of incorrect updates if an incorrect id_run is entered by the loaders e.g. missing
                  # digit so that it matches an old run which is "run canceled" such that the automatic run
                  # folder deletion process would start removing it...
```

Changes
```diff
@@ -1,5 +1,6 @@
 LIST OF CHANGES
 
+ - staging area daemon if the folder_name is not in the DB validate the id_run before updating the DB
```
dkj, line 3:
```suggestion
 - staging area daemon: when the folder_name is not in the DB, check the run is "run pending" for given id_run before updating the DB
```

lib/Monitor/Staging.pm
```diff
@@ -91,9 +91,14 @@ sub find_live {
...
+                  my $run_status = $run_row->current_run_status_description();
+                  if ( $run_status ne qq[run pending] ) {
+                      warn "Skipping run $run_dir, the id_run $id_run may be wrong as this run has a status of $run_status\n";
+                      next;
+                  }
                   $run_row->update({'folder_name' => $run_folder}); # or validation will fail
```
dkj, line 100:
```suggestion
                  # TODO: check (or fix) instrument (using external_name) and slot
                  $run_row->update({'folder_name' => $run_folder}); # or validation will fail
```

dkj (commented):
Adding some comments with the rationale for limiting to `run pending`.

PR #673, opened 2022-03-21
Change sequencescape urls to https.

absanger:


mgcam:
As far as I know, the task was to change the URL that is used by the client side and see whether the client side still works. I've tracked the URL used by the client side to this line of code https://github.com/wtsi-npg/npg_tracking/blob/718968a3c51b5af988164ddec40f2b784e9b701c/lib/npg/util.pm#L60, meaning that no code changes are necessary.

The changes in the pr are not relevant to the task. The changed code is not longer in use and is retained for historical reasons. Sorry for the confusion.

PR #699, opened 2023-03-28
Add a new instrument format record for NovaSeqX.

mgcam:
Update test data check ensure that the tested do not fail in presence of new formats.

Dev database data updated

Set up a dev server, make sure the UI looks OK - DONE


dkj (approved):
surprised we were missing NovaSeq in test fixtures.

PR #700, opened 2023-03-30
Correct handling of the 'cBot 2' type instrument

mgcam:
Depends on https://github.com/wtsi-npg/npg_tracking/pull/699
tested in a dev web server


dkj (approved):
Surprised we were missing NovaSeq in the test fixtures.

PR #701, opened 2023-04-03
Novaseqx UI representation

mgcam:
Built on top of https://github.com/wtsi-npg/npg_tracking/pull/700

Has to be properly rebased when the above pr is merged. Otherwise is finished, tested in the dev web server.


mgcam:
Replaced with https://github.com/wtsi-npg/npg_tracking/pull/702

PR #703, opened 2023-04-04
Ability to retrieve info about a NovaSeqX run.

mgcam:
Parse RunParameters.xml and RunInfo.xml to retrieve necessary information.
Added a new method - platform_NovaSeqX.
Added tests for this new method and some existing methods.


kjsanger:
Looks good to me, although you might want a second opinion for someone who looks at Illumina runfolders for frequently then me!

kjsanger (approved):

PR #706, opened 2023-04-18
No wash pending status for NovaSeqX

mgcam:


lib/npg_tracking/Schema/Result/Instrument.pm
```diff
@@ -485,7 +485,7 @@ sub status_to_change_to {
     }
     my $next_auto = $STATUS_CHANGE_AUTO{$current};
 
-    if ($self->instrument_format->model eq 'NovaSeq' &&
+    if ($self->instrument_format->model =~ /NovaSeq/sms &&
```
kjsanger, line 488: The `s` regex modifier appears twice, although the tests seem happy.

kjsanger (approved):

PR #710, opened 2023-05-02
Flag NovaSeqX flowcell as patterned

mgcam:


lib/npg_tracking/illumina/run/long_info.pm
```diff
@@ -734,19 +734,18 @@ sub is_i5opposite {
 
 =head2 uses_patterned_flowcell
 
-HiSeqX, HiSeq3000/4000, and NovaSeq use patterned flowcells
+HiSeqX, HiSeq3000/4000, and NovaSeq(X) use patterned flowcells
 https://www.illumina.com/science/technology/next-generation-sequencing/sequencing-technology/patterned-flow-cells.html
 
-Method returns true if this is one of those platforms.
+Method returns true if teh run is performed on one of those platforms.
```
dkj, line 740:
```suggestion
Method returns true if the run is performed on one of those platforms.
```

mgcam: Thanks, fixed in https://github.com/wtsi-npg/npg_tracking/pull/710/commits/d337425632cf9cc49253a72a8750fa42026f8a67

PR #712, opened 2023-05-04
Reverse complement detection

mgcam:


lib/npg_tracking/illumina/run/long_info.pm
```diff
@@ -709,6 +711,36 @@ sub is_rapid_run_abovev2 {
   return $version && ($version > 2);
 }
 
+=head2 is_read_reverse_complement
+
```
dkj, line 715: I find this method naming a bit confusing - I think its functionality is `_does_runinfo_indicate_i5_is_rev_complement` and it could be "hidden" as an attribute starting with `_` as I'd expect other code to only use `is_i5opposite` (and this "helper" to only be used in that).

dkj (approved):
This could be merged as is, as it is functionally useful now.

Not sure about the name of new attribute (see comment). Also, for consideration, is that for future proofing (not sure it is worth it) the `is_i5opposite` logic use any `IsReverseComplement` attribute found in the `Read` records in the `RunInfo.xml` as definitive irrespective of instrument type, then falls back to the older instrument "work it out" logic...

dkj (approved):

PR #718, opened 2023-05-22
Simplified run complete and mirrored detection.

mgcam:
Removed exessive complexity in finding RTACpmplete.txt and CopyComplete.txt files.

Removed a second round of examining the above files at the stage of deciding whether the mirroring has been completed.

Removed examination of the content of the Error log file since it is not present in the runfolders created by the contemporary instruments.

### DIFF
--- a/bin/staging_area_monitor
+++ b/bin/staging_area_monitor
@@ -135,8 +135,7 @@ sub check_path { ##### Function to deal with one run folder
             $folder->update_run_record(); # Inspect and update cycles,
                                           # including expected cycle count!
 
-            if ( $folder->mirroring_complete($run_path) &&
-                 $folder->check_tiles($run_path) ) {
+            if ( $folder->check_tiles($run_path) ) {
 
                 my $previous_size = $previous_size_of->{$run_path};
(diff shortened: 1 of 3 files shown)


dkj (approved):
These changes look good to me.

~Reviewing this code I'm surprised to find the file size check change is for `run complete` rather than `run mirrored`. If that's the case we may want (not high priority) to rejig that (especially when considering `/Analysis/` ouput from on-board DRAGEN).~

[would be good to run against dev DB on some live runs - with user that cannot write to run folders...]

mgcam:
On 23/05/2023 08:45, dkj wrote:
>
> Reviewing this code I'm surprised to find the file size check change 
> is for |run complete| rather than |run mirrored|. If that's the case 
> we may want (not high priority) to rejig that (especially when 
> considering |/Analysis/| ouput from on-board DRAGEN).
>
I think this check is called by the script at the stage when the run is 
already marked as complete. So seems to be correct.
>
> Message ID: ***@***.***>
>


-- 
 The Wellcome Sanger Institute is operated by Genome Research 

PR #721, opened 2023-06-01
Introduced a direct dependency on mlwh ORM.

mgcam:
This dependency was previously hidden, rather than non-existent since in the context outside of the pipeline the multi-LIMS warehouse database is the only source of LIMS data.


mgcam:
Replaced by https://github.com/wtsi-npg/npg_tracking/pull/723

PR #726, opened 2023-06-02
 Fixed a bug, an incorrect method name.

mgcam:
Staging area daemon script: platform_is_NovaSeqX() -> platform_NovaSeqX()


dkj (approved):
Runs with dev DB on live fs okay

PR #735, opened 2023-06-23
Samplesheetfor NovaSeq X Series

mgcam:


dkj:
Using already with value.

dkj (approved):
Using successfully - valuable as it is for sure.

lib/npg/samplesheet/novaseq_xseries.pm
```diff
@@ -0,0 +1,905 @@
...
+sub add_bclconvert_section {
+  my $self = shift;
+
+  my ($index1_length, $index2_length) = @{$self->_index_reads_length()};
+
+  my @lines = ();
+  push @lines, ['[BCLConvert_Settings]'];
+  push @lines, [q[SoftwareVersion], $SOFTWARE_VERSION];
+
+  # Not clear what CLI analysis option thie coresponds to.
```
jmtcsngr, line 511:
```suggestion
  # Not clear what CLI analysis option this corresponds to.
```

lib/npg/samplesheet/novaseq_xseries.pm
```diff
@@ -0,0 +1,905 @@
...
+  # Looks likely to be a list of lanes to run a tag collision check.
+  # According to @srl, bcl-covert tries to correct one error by default
+  # but it checks the tags allow this I.e that they all differ by at least
```
jmtcsngr, line 514:
```suggestion
  # but it checks the tags allow this i.e. that they all differ by at least
```

lib/npg/samplesheet/novaseq_xseries.pm
```diff
@@ -0,0 +1,905 @@
...
+  # $add_line->(qw(AdapterRead1 SOME));
+  # $add_line->(qw(AdapterRead2 OTHER));
+
+  push @lines, [];
+  push @lines, ['[BCLConvert_Data]'];
+
+  my @data_header = qw(Lane Sample_ID);
+  if ($index1_length) {
+    push @data_header, q[Index];
+    if ($index2_length) {
```
jmtcsngr, line 550: From `_index_reads_length` I understood is possible to get index2_length even if index1_length is not there. Should we be paranoid and check? Guess it may mean a problem. Maybe register issue and add in next version?

jmtcsngr (commented):
Two typos I think and a question.

lib/npg/samplesheet/novaseq_xseries.pm
mgcam: I am trying not to do too much since we are still finding out what it should be. At the end the code will have to be further refactoted and supplied with proper tests. Then we can decide how paranoid we want to be. Crasy configurations should be rejected by Illumina software.

PR #739, opened 2023-07-03
Fixed a regression in npg_move_runfolder.

mgcam:
MooseX::Getopt is no longer used by the underlying class, therefore the arguments have to be processed explicitly. With this change a better control of
input arguments is achieved.

Documentation is updated and extended.


dkj (approved):
help will be much more useful!

PR #740, opened 2023-07-04
Stop using XML LIMS data in a test.

mgcam:
Use mlwarehouse LIMS driver and data for the transcriptoms test.


dkj (approved):
I find it's much clearer using the MLWH, than the XML hacks.

PR #744, opened 2023-07-11
locate_runfolder: simplify, cope with complex globs, link page for multiple locations

dkj:
Cope with `{a,b}*` in glob.
Some vestigial functionality is removed.
Present page with options when there are multiple results from ~the~ [any] glob.


mgcam:
We've seen this script working correctly

PR #746, opened 2023-07-20
Added tests for top level staging monitor.

mgcam:


nerdstrike (changes requested):
Seems legit. Couple of inconsequential corrections to make

nerdstrike (approved):

PR #764, opened 2023-10-18
update httpd.conf with new values for mpm event

jmtcsngr:
Previous values were causing issues in Jammy. New values use more resources but service is more stable.

Originally found at:

https://dev.to/joetancy/php-fpm-with-apache2-2mk0
https://thecustomizewindows.com/2023/04/how-to-install-apache-mpm-worker-and-php-fpm-on-ubuntu-server/

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,7 @@
 LIST OF CHANGES
 
+ - update httpd.conf with new values for mpm event
+
 release 97.0.0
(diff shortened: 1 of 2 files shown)


jmtcsngr:
Most likely to be replaced by https://github.com/wtsi-npg/npg_tracking/pull/774

PR #766, opened 2023-10-23
Updated DBIx classes from the prod database.

mgcam:
Changes are due to the MySQL database server upgrade to v8.+


nerdstrike (approved):
The MySQL docs seem to imply that `RESTRICT` and `NO ACTION` are for the most part equivalent. Still, if this is what DBIC+mysql wants to do, then so be it.

PR #772, opened 2023-10-30
Added a method for aggregation by library

mgcam:


dkj (approved):
Code eye-balled and together with tests here and use of change with downstream library/tools all suggest this is good.

The aggregation method is conservative/safe - we could potentially relax contraints _in the future_ such as the mentioned run level (but also common tag_index, common set across lanes).

PR #774, opened 2023-11-01
Single change event cfg so we can bugfix release

jmtcsngr:
Pull out specific config change bits from the other [PR](https://github.com/wtsi-npg/npg_tracking/pull/764) to create a bugfix release.


mgcam (approved):
Already deployed in production, seems to have resolved problems with frequent timeouts.

PR #777, opened 2023-11-17
Distinguish between NovaSeq and NovaSeqX RunParams.

mgcam:
We use the name of the Application in RunParameters.xml to detect that the instrument belongs to the NovaSeq type. Newer NovaSeqX instrument we have has the following definition for the application:
  <Application>NovaSeqXSeries Control Software</Application>
This pull request fixes the reg exp we use to ensure that the above definition does not result in the instrument being classified as NovaSeq.

### DIFF
--- a/lib/npg_tracking/illumina/run/long_info.pm
+++ b/lib/npg_tracking/illumina/run/long_info.pm
@@ -592,7 +592,7 @@ instrument belonging to NovaSeq platform.
 
 sub platform_NovaSeq {
   my $self = shift;
-  return $self->_software_application_name() =~ /NovaSeq/xms;
+  return $self->_software_application_name() =~ /NovaSeq(?!X)/xms;
 }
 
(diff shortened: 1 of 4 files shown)


lib/npg_tracking/illumina/run/long_info.pm
```diff
@@ -592,7 +592,7 @@ instrument belonging to NovaSeq platform.
 
 sub platform_NovaSeq {
   my $self = shift;
-  return $self->_software_application_name() =~ /NovaSeq/xms;
+  return $self->_software_application_name() =~ /NovaSeq(?!X)/xms;
```
dkj, line 595:
```suggestion
  return not $self->platformNovaSeqX and $self->_software_application_name() =~ /NovaSeq[ ]/xms;
```
(following office discussion of alternatives here)

dkj (approved):

PR #778, opened 2023-11-22
Ensure insert size is returned for compositions.

mgcam:


lib/st/api/lims.pm
```diff
@@ -498,7 +498,7 @@ sub _build_required_insert_size {
   my $self = shift;
 
   my $is_hash = {};
-  if (defined $self->position) {
+  if (defined $self->position or defined $self->rpt_list) {
```
dkj, line 501: Curious, do you know if we could we drop the condition altogether?
```
  if (defined $self->position or defined $self->rpt_list) {
```

dkj (approved):

lib/st/api/lims.pm
mgcam: The only purpose of this condition, which appears in many other places in the code, is to supress a meaningless computation and return for run-level objects. As this case shows, as more entry points are added (rpt_list), one has to revisit all these conditions.

It did cross my mind to set a lazy-built `is_run_level` or similar attribute, simplify the code and make if future-proof. I am not sure that dropping this condition alltogether is helpful. There might be tests in place that demonstrate that on run level some methods are not meant to return anything. Now they will return something nobody is interested in or the methods might even error. Removing the run-level tests will be the only reasonable option, but then an unsuspecting developer might ask why there are no tests

dkj: This seems big - let's leave for another day!

PR #779, opened 2023-11-24
Get DRAGEN version for NovaSeqX samplesheet from instruments' records

mgcam:


dkj (approved):
The DRAGEN versions in the fixture almost tripped me up reviewing - but nicely test that the latest is found :+1: 

[and samplesheet generated correctly in my hands]

PR #794, opened 2023-12-15
Improved samplesheet generation for NovaSeqX.

mgcam:
Commented out code deleted.
Attributes and methods are ordered more logically. Some private accesors are converted to public.
Documentation improved.
Added index_read_length and read_length attributes so that these values can be set by the caller.

### DIFF
--- a/lib/npg/samplesheet/novaseq_xseries.pm
+++ b/lib/npg/samplesheet/novaseq_xseries.pm
@@ -201,7 +201,7 @@ has 'read_length' => (
   'is'         => 'ro',
   'lazy_build' => 1,
   'required'   => 0,
-  'documentation' => 'An array containing the length of the forward and the ' .
+  'documentation' => 'An array containing the length of the forward and ' .
                      'reverse read',
 );
 sub _build_read_length {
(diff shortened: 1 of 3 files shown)


lib/npg/samplesheet/novaseq_xseries.pm
```diff
@@ -158,9 +158,60 @@ sub _build_batch_id {
...
+
+=cut
+
+has 'index_read_length' => (
+  'isa'        => 'ArrayRef',
+  'is'         => 'ro',
+  'lazy_build' => 1,
+  'required'   => 0,
+  'documentation' => 'An array containing the length of the first and the ' .
+                     'second (if applicable) indexing read..',
```
jmtcsngr, line 178:
```suggestion
                     'second (if applicable) indexing read.',
```

lib/npg/samplesheet/novaseq_xseries.pm
```diff
@@ -158,9 +158,60 @@ sub _build_batch_id {
...
+
+An array containing the length of the first and the second (if applicable)
+indexing read.
+
+If not set, is computed as the longest first and second barcode as reported
+by the LIMS system.
+
+=cut
+
+has 'index_read_length' => (
```
jmtcsngr, line 172: Is this meant to be used as an override? Should we care if only one is provided at executing time but there are two lengths in LIMS?

jmtcsngr (commented):
Only suggestions and a couple of questions.

lib/npg/samplesheet/novaseq_xseries.pm
mgcam: This is when we want to do something semi-manually. Prior to the start of the run we only have LIMS information. If we want to run DRAGEN off-board, then we will find out what happenned from RunInfo.xml As we know only two well, all sorts of discrepancies are possible.

PR #795, opened 2023-12-15
Ability to set a path for NovaSeqX samplesheet

mgcam:
Built on top of https://github.com/wtsi-npg/npg_tracking/pull/794

Created a parent class for two existing samplesheet
generators. Moved common methods to this class.

Through this inheritance the NovaSeqX samplesheet
generator gets a standard method for setting a path for the
samplesheet.

An option to set the path from a value in the configuration file
is dropped - it was not used for years.


jmtcsngr (commented):
Small changes and suggestions.

PR #796, opened 2023-12-18
Daemon4novaseqx samplesheet

mgcam:
Built on top of https://github.com/wtsi-npg/npg_tracking/pull/795

### DIFF
--- a/lib/npg/samplesheet/auto.pm
+++ b/lib/npg/samplesheet/auto.pm
@@ -156,8 +156,7 @@ sub process {
       $ss = npg::samplesheet::novaseq_xseries->new(
         run => $r, id_run => $id_run,
         mlwh_schema => $self->mlwh_schema,
-        align => 1, keep_fastq => 1,
-        varcall => q(AllVariantCallers)
+        align => 1, keep_fastq => 1
       );
     }
 
--- a/lib/npg/samplesheet/auto.pm
+++ b/lib/npg/samplesheet/auto.pm
@@ -155,7 +155,6 @@ sub process {
       );
     } else {
       $ss = npg::samplesheet::novaseq_xseries->new(
-        id_run => $id_run,
         run => $r,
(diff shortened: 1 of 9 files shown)


lib/npg/samplesheet/auto.pm
```diff
@@ -141,9 +147,21 @@ sub process {
...
-    );
+    my $ss;
+    if ($self->instrument_format eq $MISEQ_INSTRUMENT_FORMAT) {
+      $ss = npg::samplesheet->new(
+        run => $r,
+        mlwh_schema => $self->mlwh_schema
+      );
+    } else {
+      $ss = npg::samplesheet::novaseq_xseries->new(
+        id_run => $id_run,
```
jmtcsngr, line 158: Starting to feel wary about passing id_run and run in the constructor. It will be wasteful to let the lazy build query again to get the run. As I understand we also need id_run because that way things are compatible when passing from command line. Don't think it is wrong, it just feels strange. Maybe there is something else I'm missing.

jmtcsngr (commented):
A small thing in docs and a question regarding the building of an object.

lib/npg/samplesheet/auto.pm
mgcam: Fixed, see https://github.com/wtsi-npg/npg_tracking/pull/796/commits/9a2c9f0a29aeab37388b0716af7d057aaedea986

lib/npg/samplesheet/auto.pm
```diff
@@ -8,18 +8,24 @@ use Readonly;
 use File::Copy;
 use File::Spec::Functions;
 use Carp;
+use List::MoreUtils qw(none);
```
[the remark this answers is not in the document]
mgcam: Fixed, see Fixed, see https://github.com/wtsi-npg/npg_tracking/pull/796/commits/9a2c9f0a29aeab37388b0716af7d057aaedea986

PR #798, opened 2024-01-04
Removed st::api::lims batch_id constructor arg.

mgcam:
Was needed when we used the xml driver.
No current lims driver is using this argument. Existing production code does not pass batch_id to st::api::lims constructor.

st::api::lims documentation updated.

### DIFF
--- a/lib/st/api/lims.pm
+++ b/lib/st/api/lims.pm
@@ -59,8 +59,8 @@ database. If the mlwh_schema constructor argument is not set, a connection
 to the database defined in a standard NPG configuration file is be used.
 
 Any driver attribute can be passed through to the driver's constructor via
-this objects's constructor. Not all of the attributes passed through to the
-driver are be available as this object's attributes. Example:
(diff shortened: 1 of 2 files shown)


lib/st/api/lims.pm
```diff
@@ -21,27 +21,46 @@ st::api::lims
...
+The default driver type is 'samplesheet'. The path to the samplesheet can be
+set either in the 'path' constructor attribute or by setting the env. variable
+NPG_CACHED_SAMPLESHEET_FILE.
+
+All flavours of the ml_warehouse driver require access to the ml warehouse
+database. If the mlwh_schema constructor argument is not set, a connection
+to the database defined in a standard NPG configuration file is be used.
+
+Any driver attribute can be passed through to the driver's constructor via
+this objects's constructor. Not all of the attributes passed through to the
```
jmtcsngr, line 62:
```suggestion
this objects' constructor. Not all of the attributes passed through to the
```

PR #806, opened 2024-01-30
Simplified validation of the run folder name.

mgcam:
Added a simple check to Monitor::Staging::find_live method.

Deleted the npg_tracking::illumina::run::folder::validation role, where this validation was previously implemented.


lib/Monitor/Staging.pm
```diff
@@ -68,55 +67,56 @@ sub find_live {
...
-                }    
-                if ( npg_tracking::illumina::run::folder::validation->new(
-                         run_folder          => $run_folder,
-                         id_run              => $id_run,
-                         npg_tracking_schema => $self->schema )->check() ) {
-                    $path_for{$id_run} = $run_dir;
-                    warn "Cached $run_dir for run $id_run\n";
+                my $db_folder_name = $run_row->folder_name;
+                if ( $db_folder_name ) {
+                    if ($db_folder_name ne $run_folder) {
```
srl147, line 72: Given we get the id_run from the runfolder_path are we ever likely to see a mismatch on the folder_name ? I guess there's no harm in checking anyway

mgcam: This is what the deleted role was doing. DJ asked to salvage anythign usable from it. You can imagine a far-fetched scenario when somebody accidentally edited the record.

PR #818, opened 2024-04-09
Stopped generating InstrumentFormat ORM class.

mgcam:
Dropped instrument_utilisation table from the db schema, which is used for testing the tracking web application.

'instrument_utilisation' table has not been used for the last seven years. The code that filled it in was deleted some time ago.

Once this change is released, the 'instrument_utilisation' table can be dropped from the prod. database.


jmtcsngr:
There is still a view for `instrument_utilisation` declared in [`data/config.ini`](https://github.com/mgcam/npg_tracking/blob/37f179e5d5ebff014639bd8fb1af8f90713c4280/data/config.ini#L4) (last item in the list). Do we need it if there is no ORM?

mgcam:
> There is still a view for `instrument_utilisation` declared in [`data/config.ini`](https://github.com/mgcam/npg_tracking/blob/37f179e5d5ebff014639bd8fb1af8f90713c4280/data/config.ini#L4) (last item in the list). Do we need it if there is no ORM?

No, but there are plenty of views there, which do not exist. This one will be yet another one.

PR #826, opened 2024-05-17
Added a default dnap-ssr email recipient

mgcam:
... to a list of recipients of emails for subscribers. This type of email is sent for every eligible event.


dkj:
I think this might not be what is wanted - querying on ticket to check - I suspect it is the content of the email they want changed...

mgcam:
Indeed, not what is wanted

PR #833, opened 2024-05-31
Provided access to the default study config.

mgcam:
Provided a direct access to the default section of the study configuration file.

### DIFF
--- a/lib/npg_tracking/util/pipeline_config.pm
+++ b/lib/npg_tracking/util/pipeline_config.pm
@@ -151,6 +151,24 @@ sub read_config {
   return $config;
 }
 
+=head2 default_study_config
+
+  Example    : $obj->study_config();
+  Description: Returns a default config, which is normally used as a fall-back
+               position when a study-specific config is not available. If the
+               default section of the study configuration file is not present,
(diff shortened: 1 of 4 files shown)


lib/npg_tracking/util/pipeline_config.pm
```diff
@@ -151,6 +151,24 @@ sub read_config {
...
+  Description: Returns a default config, which is normally used as a fall-back
+               position when a study-specific config is not available. If the
+               default section of the study configuration file is not present,
+               an empty hash is returned.
+
+  Returntype : Hash
+
+=cut
+
+sub default_study_config {
```
dkj, line 166: Should we use this method at line 222 ?

dkj (approved):
Ok - but consider using new method at line 222 in `pipeline_config.pm`

lib/npg_tracking/util/pipeline_config.pm
dkj: Resolved with https://github.com/wtsi-npg/npg_tracking/pull/833/commits/b1b3f7ec28bb2cbf76af36642d6e9cebe8b5da83

PR #853, opened 2024-10-11
Added sample_uuid, sample_lims methods to st::api::lims

mgcam:
Depends on https://github.com/wtsi-npg/ml_warehouse/pull/244

Patched st::api::lims::samplesheet to allow for scalar attributes ending with 's'. Previously the values of attributes with names ending with 's' were automatically converted to lists. In fact, this was necessary only for comma-separated strings of email addresses. The part of code that deals with email addresses was changed to work on attributes with names starting with 'email' rather than ending with 's'. Prior to applying this patch the value of the new sample_lims attribute was returned as a list when a samplesheet was used as a source of LIMS data.

Updated tests and test data.

sample_uuid and sample_lims is undefined for merged data where the attributes differ in merged components. Is there a need at this point in sample_uuids method for cases like this?


lib/st/api/lims/samplesheet.pm
```diff
@@ -315,7 +315,7 @@ for my $m ( st::api::lims->driver_method_list_short(__PACKAGE__->meta->get_attri
           if (defined $value && $value eq q[]) {
             $value = undef;
           }
-          if ($m =~ /s$/smx) {
```
dkj, line 318: Is `email_addresses` the only possible "plural" in our samplesheet (de)serialisation?

dkj: Could we add a code comment  - this is to deal with possible multiple values ?

Alternative would be `($m =~ /s$/smx && $m !~ /lims$/smx)` I suppose.

Not bothered about this alternative (unless you like it) - but think I'd find the comment useful.

dkj: what went wrong if we left it as it was? i.e. just `($m =~ /s$/smx)`

mgcam: The value of sample_lims were returned as an array when the samplesheet driver was used and as a scalar when teh mlwh driver was used.

mgcam: Some [documentation ](https://github.com/wtsi-npg/npg_tracking/pull/853/commits/c78d854e1e3bf820e7ea8e3e3cd56ecffe9163bf)is added

mgcam: One way of another the regexp has to mention one of the properties explicitly, either emails or sample_lims. I believe allowing for properties with names ending with 's' is a better option.

PR #854, opened 2024-10-16
Added is_lane and composition_object methods.

mgcam:
Added is_lane and composition_object methods to st::api::lims

### DIFF
--- a/lib/st/api/lims.pm
+++ b/lib/st/api/lims.pm
@@ -877,14 +877,21 @@ sub is_composition {
 
 =head2 composition_object
 
- Returns C<npg_tracking::glossary::composition> object for this LIMS entity.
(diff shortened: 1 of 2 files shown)


lib/st/api/lims.pm
```diff
@@ -873,6 +875,69 @@ sub is_composition {
   return $self->rpt_list ? 1 : 0;
 }
 
+=head2 composition_object
+
+ Returns C<npg_tracking::glossary::composition> object for this LIMS entity.
+ An undefined value is returned for a run-level object. For some drivers
+ the value of C<id_run> attribute might be undefined, in which case the return
+ value of this method is underined.
```
jmtcsngr, line 883:
```suggestion
 value of this method is undefined.
```

PR #863, opened 2025-01-08
manually install libmodule-runtime-perl to fix CI

jmtcsngr:
Ubuntu 24.04 is missing the package and we need it for running perl tests.


kjsanger (approved):

jmtcsngr:
Thanks @kjsanger !

PR #871, opened 2025-01-23
Drop stuck runs menu

mgcam:
Built on top of https://github.com/wtsi-npg/npg_tracking/pull/870


marcomoscasgr (approved):
Changes applied to a test tracking server on a VM and working. Stuck runs menu is no longer available.

PR #885, opened 2025-05-08
Create db records for AVITI instruments

mgcam:


nerdstrike (approved):
I approve on the basis that this was used to get your trials for Elembio instruments going.

PR #886, opened 2025-05-13
Elembio tracking daemon

mgcam:
Built on top of https://github.com/wtsi-npg/npg_tracking/pull/884

### DIFF
--- a/lib/npg_tracking/daemon/elembiostaging.pm
+++ b/lib/npg_tracking/daemon/elembiostaging.pm
@@ -1,0 +1,94 @@
+package npg_tracking::daemon::elembiostaging;
+
+use Moose;
+use namespace::autoclean;
+use Readonly;
+
+extends 'npg_tracking::daemon';
+
+our $VERSION = '0';
+
+Readonly::Scalar my $SCRIPT_NAME => q[elembio_staging_area_monitor];
+Readonly::Array  my @STAGING_AREAS => qw(/lustre/scratch120/elembio/staging);
+
+override 'daemon_name'  => sub { return $SCRIPT_NAME; };
+override 'command'      => sub {
+  ## no critic (CodeLayout::ProhibitParensWithBuiltins)  
+  return sprintf '%s %s',
+                 $SCRIPT_NAME,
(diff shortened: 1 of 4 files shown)


lib/npg_tracking/daemon/elembiostaging.pm
```diff
@@ -0,0 +1,94 @@
...
+
+Readonly::Scalar my $SCRIPT_NAME => q[elembio_staging_area_monitor];
+Readonly::Array  my @STAGING_AREAS => qw(/lustre/scratch120/elembio/staging);
+
+override 'daemon_name'  => sub { return $SCRIPT_NAME; };
+override 'command'      => sub {
+  ## no critic (CodeLayout::ProhibitParensWithBuiltins)  
+  return sprintf '%s %s',
+                 $SCRIPT_NAME,
+                 join(q[ ], map { "--staging_area $_"} @STAGING_AREAS);
```
marcomoscasgr, line 19: I would also add `--verbose` to get messages at RunFolder logger level

t/16-npg_tracking-daemon-elembiostaging.t
```diff
@@ -0,0 +1,14 @@
+use strict;
+use warnings;
+use Test::More tests => 4;
+
+use_ok('npg_tracking::daemon::elembiostaging');
+my $d = npg_tracking::daemon::elembiostaging->new();
+isa_ok($d, 'npg_tracking::daemon::elembiostaging');
+is ($d->daemon_name, 'elembio_staging_area_monitor', 'daemon name is correct');
+is ($d->command,
+  'elembio_staging_area_monitor --staging_area /lustre/scratch120/elembio/staging',
```
marcomoscasgr, line 10: same as above. It's better to add `--verbose` option if this will be used in production

marcomoscasgr (changes requested):

lib/npg_tracking/daemon/elembiostaging.pm
marcomoscasgr: my bad, the current behaviour will use more verbosity anyway with no option specified

marcomoscasgr (approved):

PR #893, opened 2025-05-19
Feature: elembio staging requirements

marcomoscasgr:
The new requirements will add:
1. `batch_id` parser from run name
2. `is_paired` set to 1 according to the number of sequencing reads
3. Set tags (at run creation): `staging`, `paired_read` and `multiplex`
4. New run and lane creation into a transaction
5. Assign `run stopped early` status rather than `run complete` status if the run had Outcome failed. If the run is in one of `run stopped early`, do not proceed.


lib/Monitor/Elembio/RunFolder.pm
```diff
@@ -145,21 +153,22 @@ sub _build_tracking_run {
...
+      batch_id             => $self->batch_id,
     };
-    $run_row = $rs->create($data);
-    $self->info('Created run ' . $run_row->folder_name . ' with ID ' . $run_row->id_run);
 
-    my $tag = 'staging';
-    $run_row->set_tag($USERNAME, $tag);
-    $self->info("$tag tag is set");
+    my $transaction = sub {
+      $run_row = $rs->create($data);
```
mgcam, line 161: $run_row value is reassigned here

lib/Monitor/Elembio/RunFolder.pm
```diff
@@ -427,30 +523,110 @@ sub _build__run_params_data {
...
   my $current_run_status_obj = $run_row->current_run_status;
   my $run_uploaded_path = catfile($self->runfolder_path, $RUN_UPLOAD_FILE);
-  my $is_run_complete = -e $run_uploaded_path;
   
   if ( ! $current_run_status_obj ) {
     $run_row->set_instrument_side($self->instrument_side, $USERNAME);
     $current_run_status_obj = $run_row->update_run_status($RUN_STATUS_INPROGRESS, $USERNAME, $self->date_created);
     $self->info('New run ' . $self->runfolder_path . ' created');
+    $self->_set_tags();
   }
```
mgcam, line 616: All early returns should happen before updating the number of cycles. Cases for an early return:
1. current run status is $RUN_STATUS_COMPLETE or any status after it
2. current status is one of 'run stopped early' or 'run calcelled', the latter can be set by the user via the web interface

lib/Monitor/Elembio/RunFolder.pm
```diff
@@ -427,30 +523,110 @@ sub _build__run_params_data {
...
+    if ( $current_run_status_obj->run_status_dict->compare_to_status_description($RUN_STATUS_STOPPED) == -1 ) {
+      my $date = DateTime->from_epoch(epoch => (stat  $run_uploaded_path)[9]);
+      $run_row->update_run_status($RUN_STATUS_STOPPED, $USERNAME, $date);
+      $self->info('Run ' . $self->runfolder_path . ' is failed');
+    } else {
+      $self->info('Run ' . $self->runfolder_path . ' was failed, current status ' . $current_run_status_obj->description);
+    }
+  }
+
+  if ($self->is_completed) {
```
mgcam, line 629: add to the condition: or if the run is marked as failed, but the expected number of cycles was performed

mgcam (changes requested):
In general having lots of constants does not help comprehension. If they are used only once, those that are different keys can as well be hardcoded.

As discussed earlier, using constants in test code might mask errors in the code that is being tested.

I am not asking to remove constants, this comment is for the future.

Some changes are suggested. The early returns logic should be more explicit.

lib/Monitor/Elembio/RunFolder.pm
```diff
@@ -145,21 +153,22 @@ sub _build_tracking_run {
...
-      $self->info("Created record for lane $lane of run_id " . $run_row->id_run);
-    }
+      my $runlane_rs = $run_row->result_source()->schema()->resultset($RUNLANE_TABLE);
+      for my $lane (1 .. $self->lane_count) {
+        $runlane_rs->create({id_run => $run_row->id_run, position => $lane});
+        $self->info("Created record for lane $lane of run_id " . $run_row->id_run);
+      }
+      return $run_row;
+    };
+    $run_row = $rs->result_source()->schema()->txn_do($transaction);
```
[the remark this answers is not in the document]
marcomoscasgr: Yes, after removing the reassignments it works

mgcam (approved):
Tested against dev database

PR #896, opened 2025-05-28
Added a method for finding db run records

mgcam:


t/12-dbic-ResultSet-Run.t
```diff
@@ -0,0 +1,45 @@
...
+  
+  my $irs = $schema->resultset('Instrument');
+  $irs->search({name => 'NVX1'})->next()->update({external_name => 'LH00210'});  
+  throws_ok { $rs->find_with_attributes(
+    '20250127_AV244103_NT1850075L', '2427452508', 'LH00210') }
+    qr/Multiple instrument records with name or external name LH00210/,
+    'error if multiple instruments with a given external name are present';
+
+  is ($rs->find_with_attributes(
+    '20250126_AV244103_NT1850075L', '2427452509', 'AV244103'), undef,
```
nerdstrike, line 37: I think this is stringy-comparision of "" with "", rather than testing undef. Perhaps should be `ok (defined($rs->find_with_attributes(..)), 'an undefined value..');` ?

nerdstrike (commented):
Seems to be what is needed. One question about testing undef.

nerdstrike (approved):

PR #900, opened 2025-06-16
Feature: find id run

marcomoscasgr:
Add a script to find an `id_run` given a `run folder path`.

### DIFF
--- a/bin/elembio_find_id_run
+++ b/bin/elembio_find_id_run
@@ -1,0 +1,99 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+use FindBin qw($Bin);
+use lib ( -d "$Bin/../lib/perl5" ? "$Bin/../lib/perl5" : "$Bin/../lib" );
+use English qw(-no_match_vars);
+use Readonly;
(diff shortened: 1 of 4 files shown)


mgcam (changes requested):

bin/elembio_find_id_run
```diff
@@ -0,0 +1,143 @@
...
+      $log_config = get_log_config('INFO');
+    }
+    Log::Log4perl::init(\$log_config);
+  }
+  my $log = Log::Log4perl->get_logger('main');
+
+  my $monitor = Monitor::Elembio::RunFolder->new( runfolder_path      => $runfolder_path,
+                                                  npg_tracking_schema => npg_tracking::Schema->connect());
+  $log->info('Run folder of interest: ' . $runfolder_path);
+  $log->debug('Current status of the run: ' . $monitor->tracking_run()->current_run_status_description);
```
mgcam, line 65: tracking_run is going to create a run if it does not exist, isn't it? Not what we want here.

mgcam (approved):

PR #916, opened 2025-08-08
Add tracking runs with no run type

marcomoscasgr:
15 (of the oldest) runs from AVITI23 done in 2023 have no `RunType` field in the RunParameters.json and so they were never registered in tracking. With a recent change #910 these runs will be considered but skipped again by logging an error.
This fix will consider those runs through the normal workflow as in the `Sequencing` type and register them into tracking.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,8 @@
 LIST OF CHANGES
 
+release 105.6.1 (2025-08-12)
+ - Bug fix - allow for registering old Elembio runs with no run type defined in RunParameters.json
+
 release 105.6.0 (2025-08-06)
  - Registration of 'cytoprofiling' runs in tracking DB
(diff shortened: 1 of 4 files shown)


mgcam (dismissed):
I believe it is wrong to assign sequencing type where run type is not specified. The correct action is to have an undefined value. We do not have to dance around very old data to the detriment of future data where suddenly absence of this particular definition will mean something else entirely.

marcomoscasgr:
Update on `cytoprofiling`: The `cytoprofiling` tag is not showed on the webpage, although it is in the tracking db assigned to runs.

marcomoscasgr:
Update on `batch_id`: `batch_id` is fixed now and it will not be considered for these type of runs without `RunType`

PR #962, opened 2026-03-12
Added st::api::lims driver for Elembio data.

mgcam:
Built on top of https://github.com/wtsi-npg/npg_tracking/pull/957


dkj:
This is useful as is. We may, of course, need to alter as we learn more about running elembio or ultima.

PR #965, opened 2026-03-15
Add Elembio, and generic, runfolder and lims shim classes

dkj:
Add the required runfolder and lims querying methods (in new classes) to support npg_seq_pipeline running analyses, see https://github.com/wtsi-npg/npg_seq_pipeline/pull/918


mgcam:
Extending `st::api::lims` is a separate concern and should be in a separate pr. However, I question why a generic one is needed. In the context of the pipeline we always know the manufacturer and can use an appropriate driver. If a different starting point (a lims reference) is needed, either the existing drivers can be extended or new ones can be build on the top of existing.

lib/st/api/lims/ml_warehouse_flowcell/ultima.pm
```diff
@@ -0,0 +1,359 @@
...
+);
+sub _build__wafer_rows_cache {
+  my $self = shift;
+
+  $self->has_id_flowcell_lims or
+    croak 'id_flowcell_lims should be defined';
+
+  return [
+    $self->mlwh_schema->resultset('UseqWafer')->search(
+      {batch_for_opentrons => $self->id_flowcell_lims},
```
dkj, line 129:
```suggestion
      {id_wafer_lims => $self->id_flowcell_lims},
```

lib/st/api/lims/ml_warehouse_flowcell.pm
```diff
@@ -0,0 +1,269 @@
...
+  },
+  q[Element Biosciences] => {
+    driver_class => q[st::api::lims::ml_warehouse_flowcell::elembio],
+    resultset    => q[EseqFlowcell],
+    query_column => q[id_flowcell_lims],
+  },
+  q[Ultima Genomics] => {
+    driver_class => q[st::api::lims::ml_warehouse_flowcell::ultima],
+    resultset    => q[UseqWafer],
+    query_column => q[batch_for_opentrons],
```
dkj, line 36:
```suggestion
    query_column => q[id_wafer_lims],
```

dkj (commented):
Hmmm - pretty sure I have the wrong column being used for Ultima - I need to fix or just remove the Ultima example...

dkj:
> Extending `st::api::lims` is a separate concern and should be in a separate pr. 

`lims` and `runfolder` are together as the combination of changes I needed for https://github.com/wtsi-npg/npg_seq_pipeline/pull/918 - they can be separated. (But meanwhile  I'll fix the title and description)

> However, I question why a generic one is needed. In the context of the pipeline we always know the manufacturer and can use an appropriate driver. If a different starting point (a lims reference) is needed, either the existing drivers can be extended or new ones can be build on the top of existing.

I think you're right, I think we should know the manufacturer and so be able to identify a lims driver to use, however:
- the `id_flowcell_lims` / `id_wafer_lims` should be unique for our Illumina/Elembio/Ultima runs so we should not need to use this extra query datum when asking for LIMS info (my original motivation I recall)
- I'd need to have some mapping of manufacturer to lims driver in npg_seq_pipeline (not sure I prefer that)

The main motivation for the these lims shim driver additions was to be able to work purely from LIMS info, i.e. before NPG has populated any `*_product_metrics` or other tables in MLWH (as pipeline may need to run before that happens).

[edit: also a second look after a break makes me think there is quite a lot of refactoring possible in this PR. Flipping back to draft and :pray: @mgcam ]

mgcam:
> The main motivation for the these lims shim driver additions was to be able to work purely from LIMS info, i.e. before NPG has populated any `*_product_metrics` or other tables in MLWH (as pipeline may need to run before that 

Yes, that would be very useful. However, I am not sure this code achieves this goal since it still links into product metrics tables to get tag indexes. Probably there is a default to use sequential numbers, which I have not noticed. However, (1) sequential numbers will be inconsistent with the tag indexes we are generating now and (2) since for `eseq` and `useq` controls are not represented in the LIMS part, we will have extra fun generating deplexing metrics. I should look at the Elembio pipeline results to see how this is resolved.

dkj:
> Extending `st::api::lims` is a separate concern and should be in a separate pr.

https://github.com/wtsi-npg/npg_tracking/pull/973

PR #968, opened 2026-04-02
Okta implementation with OIDC setup

avnishsinghh:


dkj:
non-blocking suggestions:
- there's use of "okta" in file names which is probably more "oidc"
- OIDCProviderMetadataURL value could be parameterised like  OIDCClientID (it's the only place Okta is actually hardcoded, right? We can provide that url in the same way we provide the ClientID)

[edit: so we could flip to another OIDC provider more easily]

avnishsinghh:
> non-blocking suggestions:
> 
> * there's use of "okta" in file names which is probably more "oidc"
> * OIDCProviderMetadataURL value could be parameterised like  OIDCClientID (it's the only place Okta is actually hardcoded, right? We can provide that url in the same way we provide the ClientID)
> 
> [edit: so we could flip to another OIDC provider more easily]

Thanks for the suggestions. I think it make sense if in future we could move to another OIDC provider. I will implement it.

wtsi_local/httpd.conf
```diff
@@ -246,6 +246,7 @@ Listen ${NPG_TRACKING_SERVER_SSL_PORT}
     </IfModule>
     DEFINE VH_SERVER_PORT ${NPG_TRACKING_SERVER_SSL_PORT}
     Include ${NPG_TRACKING_SERVER_PATH}/wtsi_local/httpd_sfweb.conf
```
marcomoscasgr, line 248: Is this still needed?

data/templates/actions.tt2
```diff
@@ -36,8 +36,8 @@
   <div class="lfloat"><img src="/icons/[% icon_prefix %]logo.png" alt="[% icon_prefix ? 'Welcome Sanger Institute' : 'Organisation' %] logo" /></div>
   <div class="rfloat">
     <div class="npg_title [% dev ? 'dev_highlight' : '' %]">[% npg_title %]</div>
-    <div><a href="/perl/enigmatic.cgi"><img src="/icons/navigation/[% image %].gif" alt="[% image == 'logout' ? 'Logout' : 'Login';  %] button" /></a>
```
marcomoscasgr, line 39: Is this `enigmatic.cgi` needed for something else?

lib/npg/authentication/sanger_oidc.pm
```diff
@@ -0,0 +1,209 @@
+package npg::authentication::sanger_oidc;
+
+use strict;
+use warnings;
+
+our $VERSION = '0.01';
```
marcomoscasgr: This variable version should be `0`. All version variables are replaced during the build.

MANIFEST
```diff
@@ -379,7 +382,7 @@ t/00-pod.t
 t/00-podcoverage.t
 t/00-scripts_compile.t
 t/10-authentication-sanger_ldap.t
```
marcomoscasgr: Is this still needed?

lib/npg/view.pm
```diff
@@ -30,10 +30,15 @@ sub new {
...
   if (!$username) {
-    my $cookie = $cgi ? $cgi->cookie(sanger_cookie_name()) : q();
-    if($cookie) {
-      $username = sanger_username($cookie, $self->util()->decription_key());
-    }
+      my $user_info = npg::authentication::sanger_oidc->new();
+      my $preferred_username = $user_info->username;
+
+      Readonly my $NOT_FOUND => -1;
+      if (index($preferred_username, q{@}) != $NOT_FOUND) {
```
marcomoscasgr: This can be simplified. This entire if/then/else block can be replaced by just `$username = (split /@/smx, $preferred_username)[0];`
This could also be a method in the module `sanger_oidc.pm`

lib/npg/authentication/sanger_oidc.pm
```diff
@@ -0,0 +1,209 @@
...
+
+sub new {
+    my ($class) = @_;
+    my $self = {};
+
+    $self->{claims} = {
```
marcomoscasgr: How are these variable set? Who/What is populating the environment with those variables?

marcomoscasgr (changes requested):
We can have a discussion on the changes  for me to understand better how things work with `Okta`

wtsi_local/httpd.conf
avnishsinghh: Yes, this is main apache config file. `tracking_oidc_auth.conf` is additional config added for OIDC.

data/templates/actions.tt2
avnishsinghh: No, this is the only place `enigmatic.cgi` was used. With the removal of old login method we don't need it.

MANIFEST
avnishsinghh: This was kept after discussion with @jmtcsngr. He confirmed @mgcam is keen on keeping this for some time. Although, it is not getting used anywhere now.

lib/npg/authentication/sanger_oidc.pm
avnishsinghh: These variable are set by OIDC on successful authentication.

lib/npg/authentication/sanger_oidc.pm
avnishsinghh: Suggestion Implemented.

lib/npg/view.pm
avnishsinghh: Removed redundant code and moved  username processing to `Sanger_oidc.pm`.

cgi-bin/oidc-login-logout.cgi
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/perl
```
marcomoscasgr, line 1: To follow the indentation convention, all perl code tabs should have two spaces

cgi-bin/oidc-login-logout.cgi
```diff
@@ -0,0 +1,68 @@
...
+use Carp qw(croak);
+
+main();
+exit 0;
+
+sub main {
+        my $q = CGI->new;
+
+        my $current_url = $ENV{'HTTP_REFERER'};
+        my $username    = $ENV{'OIDC_CLAIM_preferred_username'} // 'not-defined';
```
marcomoscasgr: This can be simplified with the assignment to $ENV{...}. If empty it will return empty string and evaluated to false. At this point the next if guard can be only if($username)...

cgi-bin/oidc-login-logout.cgi
```diff
@@ -0,0 +1,68 @@
...
+        return $q->cookie(
+                -name    => $name,
+                -value   => $value,
+                -path    => q{/},
+                -secure  => 0,
+                -expires => '+1h',
+        );
+}
+
+sub clear_cookie {
```
marcomoscasgr: This function is not used in this file

lib/npg/authentication/sanger_oidc.pm
```diff
@@ -0,0 +1,217 @@
...
+        sub          => $ENV{'OIDC_CLAIM_sub'},
+        email        => $ENV{'OIDC_CLAIM_email'},
+        name         => $ENV{'OIDC_CLAIM_name'},
+        username     => $ENV{'OIDC_CLAIM_preferred_username'},
+        groups       => $ENV{'OIDC_CLAIM_groups'},
```
marcomoscasgr: Some of these OKTA variables are not used by login/logout actions or by other code, so it's better to remove them from both the code and the tests. They can be mentioned in the description of the script.

lib/npg/authentication/sanger_oidc.pm
```diff
@@ -0,0 +1,217 @@
...
+    bless $self, $class;
+    return $self;
+}
+
+#############
+# Accessors #
+#############
+
+sub subject      { return shift->{claims}->{sub}; }
+sub email        { return shift->{claims}->{email}; }
```
marcomoscasgr: This type of methods should follow the standard format:
my $self = shift;
return $self->{claims}->{email};

lib/npg/authentication/sanger_oidc.pm
```diff
@@ -0,0 +1,217 @@
...
+1;
+=pod
+
+=head1 NAME
+
+npg::authentication::sanger_oidc - Simple OIDC claim wrapper
+
+=head1 VERSION
+
+Version 0.01
```
marcomoscasgr: This line can be removed

lib/npg/authentication/sanger_oidc.pm
```diff
@@ -0,0 +1,217 @@
...
+
+=item * OIDC_CLAIM_sub
+
+=item * OIDC_CLAIM_email
+
+=item * OIDC_CLAIM_name
+
+=item * OIDC_CLAIM_preferred_username
+
+=item * OIDC_CLAIM_groups
```
marcomoscasgr: Please also remove these items when you remove them from code. You can mention that these variable exist in the description of this module.

lib/npg/view.pm
```diff
@@ -9,8 +9,7 @@ use Try::Tiny;
 use npg::util;
 use npg::model::user;
 use npg::model::usergroup;
-use npg::authentication::sanger_sso qw/sanger_cookie_name sanger_username/;
-use npg::authentication::sanger_ldap qw/person_info/;
```
marcomoscasgr, line 13: This module file npg::authentication::sanger_ldap should be removed if not used anywhere else. Together with its test file and the line from MANIFEST file.

lib/npg/view.pm
```diff
@@ -106,13 +103,11 @@ sub realname {
 }
 
 sub person {
-  my ($self, $username) = @_;
-
-  $username ||= $self->util->requestor->username();
+  my ($self) = @_;
```
marcomoscasgr: No parameter is needed by this method. This could be assigned to shift

t/10-authentication-sanger_oidc.t
```diff
@@ -0,0 +1,97 @@
+use strict;
```
marcomoscasgr, line 1: Some of the tests in this file are not needed when you will remove the unused variables

MANIFEST
marcomoscasgr: It can be removed together with the module tested

marcomoscasgr (changes requested):

data/templates/actions.tt2
marcomoscasgr, line 39: If this file is not used any longer, it should be removed

marcomoscasgr (changes requested):

cgi-bin/oidc-login-logout.cgi
avnishsinghh: Suggestion Implemented.

cgi-bin/oidc-login-logout.cgi
avnishsinghh: Removed `not-defined` and changed if condition.

Simple `if($username)` is not working since `OIDC_CLAIM_preferred_username` might not present at all if user is not logged-in which is why I added `not-defined` in first place.

cgi-bin/oidc-login-logout.cgi
avnishsinghh: Unused function clear_cookie() removed.

data/templates/actions.tt2
avnishsinghh: This file is already removed as part of change in `gitlab/npg_esa` as it is copied during installation of 'npg_sso_enigmatic` done as part of `npg_esa` ansible execution.

lib/npg/view.pm
avnishsinghh: ldap module removed along with it's test and manifest file entry

lib/npg/view.pm
avnishsinghh: Suggestion implemented.

MANIFEST
avnishsinghh: Removed as suggested.

cgi-bin/oidc-login-logout.cgi
marcomoscasgr: What is the expected behaviour if the variable doesn't exist? Should the function skip both login and logout?

avnishsinghh: Login page should be invoked if `OIDC_CLAIM_preferred_username` is not set or doesn't exists and if it has a valid value then logout page should be invoked.

avnishsinghh: Suggestion implemented.

lib/npg/authentication/sanger_oidc.pm
avnishsinghh: Suggestion Implemented.

lib/npg/authentication/sanger_oidc.pm
avnishsinghh: Additional attributes removed and info added in documentation for future use.

lib/npg/authentication/sanger_oidc.pm
avnishsinghh: Suggestion implemented.

t/10-authentication-sanger_oidc.t
avnishsinghh: Suggestion implemented.

avnishsinghh:
I have addressed all the review comments. Please let me know if you find anything else.

lib/npg/authentication/sanger_oidc.pm
```diff
@@ -0,0 +1,221 @@
...
+  my $self = shift;
+  return $self->{claims}->{name};
+}
+
+sub username {
+  my $self = shift;
+  my $preferred_username = $self->{claims}->{username};
+
+  if ($preferred_username) {
+    return (split /@/smx, $preferred_username)[0];
```
marcomoscasgr, line 31: I have spotted this in `npg_qc` too. What happens if the email is malformed in someway? The `split` function could return unexpected output.
Could you please use the following instead to match the specific part of the string?
```
my ($username) =  $preferred_username =~ /\A(\w+)@\w+/smx;
return $username;
```
In this case, no need for the `if` statement and the lone `return`

cgi-bin/oidc-login-logout.cgi
```diff
@@ -0,0 +1,54 @@
...
+use Carp qw(croak);
+
+main();
+exit 0;
+
+sub main {
+  my $q = CGI->new;
+
+  my $current_url = $ENV{'HTTP_REFERER'};
+  my $username    = $ENV{'OIDC_CLAIM_preferred_username'} || q{};
```
marcomoscasgr, line 15: Have you also tried without `|| q{}`?

marcomoscasgr (changes requested):

lib/npg/authentication/sanger_oidc.pm
marcomoscasgr: As discussed, we keep `split` function to cover the case when the preferred name could be just a username and not an email. Double return will be removed.

avnishsinghh: I checked again and realised I added `IF` block to avoid below line in error log:

`Use of uninitialized value $preferred_username in split at /software/npg/20260326/lib/perl5/npg/authentication/sanger_oidc.pm line 34.`

cgi-bin/oidc-login-logout.cgi
avnishsinghh: This additional check is to avoid any unforeseen failures. This is commonly used in `NPG`.

lib/npg/view.pm
```diff
@@ -30,10 +29,8 @@ sub new {
   }
 
   if (!$username) {
-    my $cookie = $cgi ? $cgi->cookie(sanger_cookie_name()) : q();
-    if($cookie) {
-      $username = sanger_username($cookie, $self->util()->decription_key());
-    }
+      my $user_info = npg::authentication::sanger_oidc->new();
```
marcomoscasgr: Wrong indentation

t/10-authentication-sanger_oidc.t
```diff
@@ -0,0 +1,51 @@
...
+# -----------------------
+
+isa_ok($oidc, 'npg::authentication::sanger_oidc');
+
+# -----------------------
+# Accessors
+# -----------------------
+
+is($oidc->name,     'Tiger Cat', 'name accessor works');
+is($oidc->username, 'tiger', 'username accessor works');
```
marcomoscasgr: We need to add more tests about this. Please consider also the cases when the input is only a username, an entire email or a bad-formatted email (such as starting with `@`)

marcomoscasgr (changes requested):

lib/npg/view.pm
avnishsinghh: Fixed indentation.

t/10-authentication-sanger_oidc.t
avnishsinghh: Increased test coverage.

lib/npg/authentication/sanger_oidc.pm
avnishsinghh: Lined removed as suggested.

PR #973, opened 2026-04-29
Extra lims MLWH drivers for elembio and ultimagen

dkj:
providing LIMS info from flowcell identification without product metrics

### DIFF
--- a/lib/st/api/lims/ml_warehouse_flowcell/elembio.pm
+++ b/lib/st/api/lims/ml_warehouse_flowcell/elembio.pm
@@ -1,0 +1,356 @@
+package st::api::lims::ml_warehouse_flowcell::elembio;
+
+use Moose;
+use MooseX::StrictConstructor;
+use namespace::autoclean;
+use Carp;
+use Readonly;
+
+use npg_tracking::util::types;
+use st::api::lims;
+use WTSI::DNAP::Warehouse::Schema;
+
+with qw/
+  npg_tracking::glossary::flowcell
+  npg_tracking::glossary::lane
+  npg_tracking::glossary::tag
+/;
+
+our $VERSION = '0';
+
+Readonly::Scalar my $LIMS_RESULT_CLASS =>
+  'WTSI::DNAP::Warehouse::Schema::Result::EseqFlowcell';
+Readonly::Array my @DELEGATED_METHODS =>
+  grep { $LIMS_RESULT_CLASS->can($_) }
+  st::api::lims->driver_method_list_short(__PACKAGE__->meta->get_attribute_list);
+
+=head1 NAME
+
+st::api::lims::ml_warehouse_flowcell::elembio
+
+=head1 SYNOPSIS
+
+  my $d = st::api::lims::ml_warehouse_flowcell::elembio->new(
+    id_flowcell_lims => 107441
+  );
+
+=head1 DESCRIPTION
+
+Element Biosciences flowcell-table-backed MLWH LIMS driver.
+
+=head1 SUBROUTINES/METHODS
+
+=head2 manufacturer
+
+=head2 children
+
+=head2 is_pool
+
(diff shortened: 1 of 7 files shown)


dkj:
Need to beware of "auto-tag_index" in absence of `_product_metrics` entries

lib/st/api/lims/ml_warehouse_flowcell/elembio.pm
```diff
@@ -0,0 +1,356 @@
...
+
+sub is_pool {
+  my $self = shift;
+  return ($self->has_position && (!$self->has_tag_index || !$self->tag_index)) ? 1 : 0;
+}
+
+sub is_control {
+  my $self = shift;
+  my $row = $self->_get_dbix_row();
+  return $row ? ($row->entity_type =~ /spike/xms ? 1 : 0) : undef;
```
mgcam, line 193: This is always false at the moment since controls are not implemented on the LIMS side.

lib/st/api/lims/ml_warehouse_flowcell/elembio.pm
```diff
@@ -0,0 +1,356 @@
...
+}
+
+sub qc_state {
+  my $self = shift;
+  my %values = map { $_ => 1 }
+    grep { defined }
+    map { $_->qc }
+    map { $_->eseq_product_metrics } $self->_rows;
```
mgcam, line 201: Find out what happens here if `$_->eseq_product_metrics` is undefined.

lib/st/api/lims/ml_warehouse_flowcell/elembio.pm
```diff
@@ -0,0 +1,356 @@
...
+    return $self->_get_dbix_row() ? $self->$orig() : undef;
+  };
+}
+
+sub _row_tag_index {
+  my ($self, $row) = @_;
+
+  my %tag_indices = map { $_ => 1 }
+    grep { defined && $_ != 0 }
+    map { $_->tag_index } $row->eseq_product_metrics;
```
mgcam, line 274: `$row->eseq_product_metrics` is a single row object. If it were a result set, `->all() `would have to be used. Therefor all this long computation on lines 272-280 can be replaced with

`if ($row->eseq_product_metrics) { return $row->eseq_product_metrics->tag_index;}`

The tag_index column is nullable. Not sure whether it is possible to get NULL the way we compute tag indexes, so probably better to return an undefine value and hope that the code fails somewhere.

lib/st/api/lims/ml_warehouse_flowcell/elembio.pm
```diff
@@ -0,0 +1,356 @@
...
+  my @tag_indices = sort { $a <=> $b } keys %tag_indices;
+  if (@tag_indices) {
+    @tag_indices == 1 or croak
+      'Multiple tag indices linked to a single eseq_flowcell row';
+    return $tag_indices[0];
+  }
+
+  my @lane_rows = sort {
```
mgcam, line 282: Probably the mapping of primary keys to tag indexes should be computed once as a private attribute

mgcam (changes requested):

lib/st/api/lims/ml_warehouse_flowcell/elembio.pm
```diff
@@ -0,0 +1,354 @@
...
+}
+
+foreach my $method (@DELEGATED_METHODS) {
+  around $method => sub {
+    my ($orig, $self) = @_;
+    return $self->_get_dbix_row() ? $self->$orig() : undef;
+  };
+}
+
+sub _row_tag_index {
```
mgcam, line 267: If we want to drive p4stage1 deplexing with a samplesheet derived with this driver, we need to represent controls. However, we cannot represent controls using LIMS tables only.

PR #975, opened 2026-05-01
Additional areas for Ultimagen staging daemon

mgcam:


nerdstrike (approved):
Seems ok. We should come back to this if Ultima don't come through with the configuration changes.

PR #980, opened 2026-05-13
Update the staging paths for Ultimagen instruments

marcomoscasgr:
- The external name that we use for `UG 200` instrument has a mismatch between the label attached on the instrument and the cloud records. Ultima people confirmed that the cloud records are the correct ones. Update the external name from `U10` to `U010`.

- Remove `Runs` subfolder when globbing for Ultimagen run folders since the instruments will write to `staging/V125` and `staging/U010` by the new settings.

- Remove the old `staging/Runs` folder used for 'V125' from the daemon.

Merge when the staging area is updated on the instrument

### DIFF
--- a/lib/npg_tracking/daemon/ultimagenstaging.pm
+++ b/lib/npg_tracking/daemon/ultimagenstaging.pm
@@ -11,7 +11,7 @@ our $VERSION = '0';
 Readonly::Scalar my $SCRIPT_NAME => q[ultimagen_staging_area_monitor];
 Readonly::Array  my @STAGING_AREAS =>
   map { '/lustre/scratch120/openstack/default/ultima/' . $_ }
-  qw( staging staging/V125 staging/U010 );
+  qw( staging/V125 staging/U010 );
 
 override 'daemon_name'  => sub { return $SCRIPT_NAME; };
 override 'command'      => sub {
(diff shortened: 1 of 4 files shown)


lib/npg_tracking/daemon/ultimagenstaging.pm
```diff
@@ -11,7 +11,7 @@ our $VERSION = '0';
 Readonly::Scalar my $SCRIPT_NAME => q[ultimagen_staging_area_monitor];
 Readonly::Array  my @STAGING_AREAS =>
   map { '/lustre/scratch120/openstack/default/ultima/' . $_ }
-  qw( staging staging/V125 staging/U10 );
+  qw( staging staging/V125 staging/U010 );
```
dkj: Do we need the lone `staging` ?

marcomoscasgr: We may want to leave the lone staging and update it to `staging/Runs` (at least until we move the runs, if we want to move them). The daemon needs update accordingly

marcomoscasgr: As discussed, the lone `staging` can be removed as those run statuses will no longer be changed by the daemon.

dkj (approved):
Fits my understanding of how the instruments are now writing.

PR #981, opened 2026-05-14
Add UG200 monitor support

marcomoscasgr:
Allow the staging monitor to read the instrument (external) name from the path where the instrument writes to. If no `instrument_name` is parsed out it should exit with error as the staging path is the only source where to get the name.

### DIFF
--- a/lib/Monitor/Ultimagen/RunFolder.pm
+++ b/lib/Monitor/Ultimagen/RunFolder.pm
@@ -28,8 +28,6 @@ Readonly::Scalar my $USERNAME => 'useq_pipeline';
 Readonly::Scalar my $RUN_UPLOADED_FILE => 'UploadCompleted.json';
 Readonly::Scalar my $RUN_STATUS_MIRRORED => 'run mirrored';
 
-Readonly::Scalar my $INSTRUMENT_NAME => 'V125';
-
 =head1 NAME
 
 Monitor::Ultimagen::RunFolder
@@ -61,6 +59,25 @@ has q{runfolder_path} => (
   is            => q{ro},
   required      => 1,
 );
+
+# The instrument (external) name parsed out from the
+# runfolder_path attribute.
+# Ultima does not store instrument info on disk, so 
+# we need to infer the instrument (external) name for
+# the run from the path where it writes to.
+has q{_instrument_external_name} => (
+  isa               => q{Str},
+  is                => q{ro},
+  required          => 0,
+  lazy_build        => 1,
+);
+sub _build__instrument_external_name {
+  my $self = shift;
+  my ($instrument_name) = $self->runfolder_path =~ m/staging\/(\w+)\//ms;
+  $instrument_name or croak 
+    'Failed to parse instrument name from runfolder path ' . $self->runfolder_path;
+  return $instrument_name;
+}
 
 =head2 runfolder_glob
 
@@ -113,7 +130,9 @@ sub _build_tracking_run {
   my $ultimagen_runid = $self->_get_ultimagen_run_attr('RunId');
 
   my $rs = $self->npg_tracking_schema->resultset('Run');
-  my $run_row = $rs->find_with_attributes($ultimagen_runid, $INSTRUMENT_NAME);
+  my $run_row = $rs->find_with_attributes(
+    $ultimagen_runid,
+    $self->tracking_instrument()->external_name);
   if ($run_row) {
     $self->info('Found run ' . $run_row->folder_name . ' with ID ' . $run_row->id_run);
     if ($run_row->folder_name ne $self->folder_name) {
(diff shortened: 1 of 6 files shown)


lib/Monitor/Ultimagen/RunFolder.pm
```diff
@@ -50,6 +48,32 @@ Run monitor for an Ultimagen run folder.
...
+  isa               => q{Str},
+  is                => q{ro},
+  required          => 0,
+  lazy_build        => 1,
+);
+sub _build_instrument_external_name {
+  my $self = shift;
+
+  my ($instrument_name) = 
+    $self->runfolder_path =~ m/.*\/staging\/(?:(\w+)\/){0,1}Runs\/\d+-\d+_\d+\Z/ms;
```
dkj: We need to drop `Runs` I think.

marcomoscasgr: After updating the daemon, I will need to change this PR. `Runs` will be optional in this case because we still have the original UG100 folder with it.

lib/Monitor/Ultimagen/RunFolder.pm
```diff
@@ -28,7 +28,8 @@ Readonly::Scalar my $USERNAME => 'useq_pipeline';
 Readonly::Scalar my $RUN_UPLOADED_FILE => 'UploadCompleted.json';
 Readonly::Scalar my $RUN_STATUS_MIRRORED => 'run mirrored';
 
-Readonly::Scalar my $INSTRUMENT_NAME => 'V125';
+# Instrument
+Readonly::Array my @INSTRUMENT_EXT_NAMES => qw/V125 U010/;
```
mgcam: It is strange to hard-code the expected names here while they are also available in the database.

mgcam:
https://github.com/marcomoscasgr/npg_tracking/blob/11054d341de3782e1c1df1393247f00c65b787ed/lib/Monitor/Ultimagen/Staging.pm#L41 - this comment has to be updated

lib/Monitor/Ultimagen/RunFolder.pm
```diff
@@ -50,6 +51,32 @@ Run monitor for an Ultimagen run folder.
...
+  isa               => q{Str},
+  is                => q{ro},
+  required          => 0,
+  lazy_build        => 1,
+);
+sub _build_instrument_external_name {
+  my $self = shift;
+  my $regex = join "|", @INSTRUMENT_EXT_NAMES;
+  my ($instrument_name) = $self->runfolder_path =~ m/\/($regex)\//ms
+   or croak 'Unknown instrument external name';
```
mgcam: A more informative error will be 'Failed to parse instrument name from runfolder path' . $self->runfolder_path

lib/Monitor/Ultimagen/RunFolder.pm
```diff
@@ -50,6 +51,32 @@ Run monitor for an Ultimagen run folder.
...
+has q{instrument_external_name} => (
+  isa               => q{Str},
+  is                => q{ro},
+  required          => 0,
+  lazy_build        => 1,
+);
+sub _build_instrument_external_name {
+  my $self = shift;
+  my $regex = join "|", @INSTRUMENT_EXT_NAMES;
+  my ($instrument_name) = $self->runfolder_path =~ m/\/($regex)\//ms
```
mgcam:
```
perl -le 'print q[qqq/U010/V125/nnn/U010/ddd] =~/\/(V125|U010)\//sm'
U010

perl -le 'print q[qqq/V125/nnn/U010/ddd] =~/\/(V125|U010)\//sm'
V125
```

Hmm...

```
perl -le 'print q[qqq/V125/staging/U010/ddd] =~/staging\/(V125|U010)\//sm'
U010
```

Better

mgcam (changes requested):

lib/Monitor/Ultimagen/RunFolder.pm
marcomoscasgr: An alternative would be to implement an attributes that retrieves all the available instrument (external) names from tracking db. Should I go for the database option?

mgcam: I think this would be better. If they have another one, we will not need to change this code.

lib/Monitor/Ultimagen/RunFolder.pm
```diff
@@ -170,15 +188,15 @@ has q{tracking_instrument} => (
...
   my $self = shift;
   my $rs = $self->npg_tracking_schema->resultset('Instrument');
-  my $params = {
-    external_name => $INSTRUMENT_NAME
-  };
-  my @instrument_rows = $rs->search($params)->all();
+  my @instrument_rows = grep {
+      $_->manufacturer_is_UltimaGenomics() and
+        ($_->external_name eq $self->_instrument_external_name)
+    } $rs->search({})->all();
```
mgcam: Why not to delegate the search to the database?

```
my @instrument_rows = grep {
  $_->manufacturer_is_UltimaGenomics()
} $rs->search({external_name => $self->_instrument_external_name})->all();
```

It is possible to join to the manufacturer table and to restrict the manufacturer, but then we have to know the name of the manufacturer. Practically speaking a grep on a list of one is OK.

mgcam (changes requested):

mgcam:
Run the staging monitor against ~/staging/U010. Two new runs were created

PR #987, opened 2026-06-04
Test config parsing special characters

ce10-sanger:


marcomoscasgr (approved):

mgcam:
t/.npg/t-magic-connection-special-chars has to be added to MANIFEST. I hope the rest of them are already there. MANIFEST is used when a Perl package is packed for distribution, which we do not do, but still.

ce10-sanger:
> t/.npg/t-magic-connection-special-chars has to be added to MANIFEST. I hope the rest of them are already there. MANIFEST is used when a Perl package is packed for distribution, which we do not do, but still.

Thank you Marina. Have addressed here https://github.com/wtsi-npg/npg_tracking/pull/988

PR #991, opened 2026-06-19
Reduce number of hours to wait after rta complete

avnishsinghh:


t/34-monitor-runfolder-staging.t
jmtcsngr: Not part of this change but worth correcting the comment. Could be in this PR as an extra commit or a different PR. I think this should be 

```suggestion
    $atime -= 9 * $SECONDS_PER_HOUR; # make it 9 hours ago
```
Do you agree?

avnishsinghh: Suggestion implemented.

jmtcsngr (approved):

PR #996, opened 2026-07-10
Change default manufacturer for the web pages to All

mksanger:
Element Biosciences and Ultima runs and instruments will now appear in the tabular pages of tracking.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,7 @@
 LIST OF CHANGES
 
+ - Changed the default manufacturer to All for the web pages.
+
 release 107.2.2 (2026-06-24)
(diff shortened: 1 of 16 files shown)


mgcam:
We probably could delete [this line](https://github.com/mksanger/npg_tracking/blob/c380c24200953b9186e9ba84cf948c0d5c0513fe/lib/npg/model/instrument_format.pm#L15)

mgcam (changes requested):
Dev server pages look correct, the changes are sensible. Please have a look at whether the line mentioned in the comment can be deleted. Otherwise all is fine.

mgcam (approved):

PR #997, opened 2026-07-14
Support the 'only cell paint' cytoprofiling type

marcomoscasgr:
Support the 'only cell paint' run type in the Cytoprofiling category for Elembio.


marcomoscasgr:
As discussed and confirmed, this PR can be merged

PR #1000, opened 2026-08-19
Added a script for managing run tags (adding or removing)

mgcam:


bin/npg_tag_manage
```diff
@@ -0,0 +1,103 @@
...
+
+  # Add a single tag
+  npg_tag_manage --id_run 1234 --tag 'staging' --username pipeline
+
+  # Add multiple tags, specify the action explicitly
+  npg_tag_manage --id_run 1234 --tag 'no auto' --tag 'staging' \
+    --action add --username pipeline
+
+  # Remove a single tag
+  npg_tag_manage--id_run 1234 --tag 'staging' --action remove
```
nerdstrike: Missing whitespace

bin/npg_tag_manage
```diff
@@ -0,0 +1,103 @@
...
+
+  # Remove multiple tags
+  npg_tag_manage--id_run 1234 --tag 'staging' --tag 'no auto' --action remove
```
nerdstrike: Here too

bin/npg_tag_manage
```diff
@@ -0,0 +1,103 @@
...
+
+=head1 REQUIRED ARGUMENTS
+
+ id_run - tracking run ID
+
+ tag - a single tag, might be specified several times
+
+=head1 OPTIONS
+
+ action - action to perform, on tag(s), C<add> or C<remove>, default to C<add>
```
nerdstrike: Are there any other options here? Could have been a flag like `--rm` or whatever.

t/10-npg_tracking-tag_manager.t
```diff
@@ -0,0 +1,128 @@
...
+  } qr/At least one tag should be given/,
+    'error when the tag attribute is set to an empty array';
+  throws_ok { npg_tracking::tag_manager->new(id_run=>1,tag=>'one',schema=>$schema) }
+    qr/username should be given/,
+    'error when username is not given when adding a tag';
+  throws_ok {
+    npg_tracking::tag_manager->new(id_run=>1,tag=>'one',action=>'some',schema=>$schema)
+  } qr/Invalid action 'some'/,
+    'error when unknown action is specified';
+
```
jmtcsngr, line 26: Maybe add tests for missing id_run?

t/10-npg_tracking-tag_manager.t
```diff
@@ -0,0 +1,128 @@
...
+  isa_ok ($tm, 'npg_tracking::tag_manager');
+  lives_ok {
+    $tm = npg_tracking::tag_manager->new(
+      id_run   => 1,
+      tag      => [qw/one two/],
+      action   => 'remove',
+      username => 'pipeline',
+      schema   => $schema
+    )
+  } 'object created OK';
```
jmtcsngr:
```suggestion
  } 'object created OK with multiple tags';
```

t/10-npg_tracking-tag_manager.t
```diff
@@ -0,0 +1,128 @@
...
+   )
+  } 'object created OK';
+  lives_ok {
+    npg_tracking::tag_manager->new(
+      id_run   => 1,
+      tag      => [qw/one two/],
+      username => 'pipeline',
+      schema => $schema
+   )
+  } 'object created OK';
```
jmtcsngr:
```suggestion
  } 'object created OK with multiple tags';
```

lib/npg_tracking/tag_manager.pm
```diff
@@ -0,0 +1,212 @@
...
+
+sub perform_action {
+  my $self = shift;
+
+  my $run_row = $self->schema->resultset('Run')->find($self->id_run);
+  if (!$run_row) {
+    croak sprintf 'Run id %i does not exist', $self->id_run;
+  }
+
+  for my $t (@{$self->tag}) {
```
jmtcsngr: Maybe report which tags have been added/removed? Because you may be successful with some but fail with one in the list?

jmtcsngr (commented):
Couple of suggestions

lib/npg_tracking/tag_manager.pm
mgcam: The code we use fails with inconsistent tags only. Otherwise anything goes. We can wrap this into a transaction so that in case of an error nothing is added or removed.

jmtcsngr: Transaction (all or nothing) sounds good.

